Issue #78, opened 2021-06-08
template space


a3sha2:
process different templates, native and anat spaces
- [x] anat
- [x] native
- [x] MNI2009 - default fmriprep 
- [x] MNI2006 FSL 
- [x] OASIS
- [x] NKI
- [x] PNC

Issue #68, opened 2021-05-24
atlas upgrade


a3sha2:
add subcortical!

mb3152:
here is citation:
https://www.nature.com/articles/s41593-020-00711-6

https://github.com/yetianmed/subcortex/tree/master/Group-Parcellation/3T/Cortex-Subcortex

Let's use these:

https://github.com/yetianmed/subcortex/blob/master/Group-Parcellation/3T/Cortex-Subcortex/Schaefer2018_400Parcels_17Networks_order_Tian_Subcortex_S1.dlabel.nii
https://github.com/yetianmed/subcortex/blob/master/Group-Parcellation/3T/Cortex-Subcortex/Schaefer2018_400Parcels_17Networks_order_Tian_Subcortex_S2.dlabel.nii
https://github.com/yetianmed/subcortex/blob/master/Group-Parcellation/3T/Cortex-Subcortex/Schaefer2018_400Parcels_17Networks_order_Tian_Subcortex_S3.dlabel.nii
https://github.com/yetianmed/subcortex/blob/master/Group-Parcellation/3T/Cortex-Subcortex/Schaefer2018_400Parcels_17Networks_order_Tian_Subcortex_S4.dlabel.nii

a3sha2:
@sattertt ^^

PR #86, opened 2021-06-10
Merge pull request #85 from PennLINC/main

a3sha2:
back to template for atlas


a3sha2:
different template PNC, NKI, MN6, MN9, OASIS, T1w and native 
#79 #78 #85  #68

Issue #267, opened 2021-12-11
-p Argument not loaded correctly


smeisler:
Using latest stable release (0.0.8) on singularity
**Command**:
`singularity run -B ${scratch},${bids_dir} $IMG ${output_dir}/fmriprep ${output_dir} participant --participant_label ${subject:4} -w $scratch -p acompcor -f 1 --smoothing 0`

**Error:**
  ```
  Running xcp_abcd version 0.0.8+0.g88e3ec9.dirty:
      * fMRIPrep directory path: /om4/group/gablab/data/ds003126/derivatives/fmriprep.
      * Participant list: ['047EPKL011012'].
      * Run identifier: 20211211-114648_a665f222-9149-4b16-8fe3-32e5863f62a3.

    
Process Process-2:
Traceback (most recent call last):
  File "/usr/local/miniconda/lib/python3.8/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/usr/local/miniconda/lib/python3.8/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/local/miniconda/lib/python3.8/site-packages/xcp_abcd/cli/run.py", line 485, in build_workflow
    retval['workflow'] = init_xcpabcd_wf (
  File "/usr/local/miniconda/lib/python3.8/site-packages/xcp_abcd/workflow/base.py", line 147, in init_xcpabcd_wf
    single_subj_wf = init_subject_wf(
  File "/usr/local/miniconda/lib/python3.8/site-packages/xcp_abcd/workflow/base.py", line 403, in init_subject_wf
    bold_postproc_wf = init_boldpostprocess_wf(bold_file=bold_file,
  File "/usr/local/miniconda/lib/python3.8/site-packages/xcp_abcd/workflow/bold.py", line 219, in init_boldpostprocess_wf
    """.format(regressors=stringforparams(params=params),sclver=sklearn.__version__,
  File "/usr/local/miniconda/lib/python3.8/site-packages/xcp_abcd/utils/utils.py", line 197, in stringforparams
    return bsignal
UnboundLocalError: local variable 'bsignal' referenced before assignment
```

Based on looking at the utils.py script, it looks like the -p argument isn't properly loaded as the `params` variable. Code runs fine when leaving this option blank and using the 27P default parameter set, but I'd like to use acompcor.

Thanks,
Steven

a3sha2:
can you try `xcp_abcd:unstable`?

[...]

PR #274, opened 2021-12-16
[Fix] Update confounds.py

smeisler:
Changes acompcor loading strategy to be compatible with fMRIPrep 21.0.0+ (#267)

This update makes acompcor component loading compatible with previous and newer versions of fMRIPrep

No documentation should be changed.

### DIFF
--- a/xcp_abcd/utils/confounds.py
+++ b/xcp_abcd/utils/confounds.py
@@ -64,7 +64,7 @@ def load_acompcor(confoundspd, confoundjs):
 
     WM = []; CSF = []
     for key, value in confoundjs.items():
-        if 'a_comp_cor' in key:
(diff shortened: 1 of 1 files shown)


a3sha2 (commented):
Thank you @smeisler 
is there any changes to tcompcor to avoid the same issues

smeisler:
I believe tcompcor is the same.

The new changes are basically:
`a_comp_cor` keys refers to ONLY combined WM/CSF mask
`w_comp_cor` refers to WM acompcor mask
`c_comp_cor` refers to CSF acompcor mask

Thankfully the existing code works with the JSON structure, so the only change that needed to be made was making sure the w/c_comp_cor keys were not being missed out.

PR #295, opened 2022-02-11
Update confounds.py

smeisler:
1) Make it so XCP does not crash if fewer than 5 WM or CSF acompcor components exist
2) Load all cosine regressors for acompcor (not just first 5, some runs will not even have 5 cosine regressors).


mb3152:
this is great, thank you!!!

PR #351, opened 2022-06-15
[RF] RemoveTR and Despike

mattcieslak:
This PR refactors and tests the `RemoveTR` interface and the `Despike` interface. 

For the `RemoveTR` interface
 - [x] Ensure that the output tsv is a proper fmriprep format confounds tsv
 - [x] Ensure the output tsv is indeed truncated 
 - [x] Ensure that both nifti and cifti are correctly truncated
 - [x] Rename interface inputs and outputs to be more clear
 - [x] Check whether we need wb_command to cut times off the beginning of a cifti file


xcp_d/interfaces/prepostcleaning.py
```diff
@@ -15,12 +15,14 @@ class _removeTRInputSpec(BaseInterfaceInputSpec):
                      mandatory=True,
                      desc=" either bold or nifti ")
     mask_file = File(exists=False, mandatory=False, desc="required for nifti")
-    volumes_to_drop = traits.Int(mandatory=True,
-                                 desc="number of volumes to drop from the beginning")
+    time_todrop = traits.Float(mandatory=True,
```
mattcieslak, line 18: I think we should actually keep this in an integer volumes to drop. We'll calculate this based on the TR in the main code

xcp_d/interfaces/prepostcleaning.py
```diff
@@ -15,12 +15,14 @@ class _removeTRInputSpec(BaseInterfaceInputSpec):
...
+                               desc="number of seconds to drop from the beginning")
     fmriprep_conf = File(
         exists=True,
         mandatory=False,
         desc="confound selected from fmriprep confound matrix")
+    TR = traits.Float(mandatory=True,
```
mattcieslak, line 24: If we're directly using number of volumes to drop we won't need the TR

xcp_d/interfaces/prepostcleaning.py
```diff
@@ -51,9 +53,9 @@ class removeTR(SimpleInterface):
     output_spec = _removeTROutputSpec
 
     def _run_interface(self, runtime):
-
+        volumes_to_drop = np.ceil(self.inputs.time_todrop / self.inputs.TR)
```
mattcieslak, line 56: Let's keep this outside of this function so we don't calculate it twice. We have to calculate it for the boilerplate in the init_postprocess_wf

xcp_d/interfaces/prepostcleaning.py
```diff
@@ -76,15 +78,15 @@ def _run_interface(self, runtime):
...
+            dropped_data = data[volumes_to_drop:]
             dropped_image = nb.Cifti2Image(
                 dropped_data,
                 header=bold_image.header,
                 nifti_header=bold_image.nifti_header)
 
         # It's a Nifti
         else:
-            dropped_data = data[..., self.inputs.volumes_to_drop:]
+            dropped_data = data[volumes_to_drop:]
```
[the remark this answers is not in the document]
kahinimehta: ahhh

PR #380, opened 2022-07-13
[ENH] Anatomical/Surface workflows

mattcieslak:
Picks up from #333

### DIFF
--- a/xcp_d/interfaces/c3.py
+++ b/xcp_d/interfaces/c3.py
@@ -1,0 +1,246 @@
+# -*- coding: utf-8 -*-
+"""Convert3D is a command-line tool for converting 3D images between common file formats."""
+import os
+from glob import glob
+
+from nipype.interfaces.base import (
+    CommandLineInputSpec,
+    traits,
+    TraitedSpec,
+    File,
+    SEMLikeCommandLine,
+    InputMultiPath,
+    OutputMultiPath,
+    CommandLine,
+    isdefined,
+)
+from nipype.utils.filemanip import split_filename
+import logging
+
+iflogger = logging.getLogger("interface")
+
+
+class C3dAffineToolInputSpec(CommandLineInputSpec):
+    reference_file = File(exists=True, argstr="-ref %s", position=1)
+    source_file = File(exists=True, argstr="-src %s", position=2)
+    transform_file = File(exists=True, argstr="%s", position=3)
+    itk_transform = traits.Either(
+        traits.Bool,
+        File(),
+        hash_files=False,
+        desc="Export ITK transform.",
+        argstr="-oitk %s",
+        position=5,
+    )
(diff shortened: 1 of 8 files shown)


xcp_d/interfaces/c3.py
```diff
@@ -0,0 +1,246 @@
+# -*- coding: utf-8 -*-
```
mattcieslak, line 1: Is this already in nipype? If it is, we should probably import it instead of copying the interface here

xcp_d/workflow/anatomical.py
```diff
@@ -1,784 +1,1371 @@
...
+from ..interfaces.c3 import C3d  # TM
+from ..interfaces.ants import CompositeInvTransformUtil  # TM
+from ..interfaces.workbench import (
+    ApplyWarpfield,
+    SurfaceGenerateInflated,
+    SurfaceAverage,
+)  # MB,TM
+
+
+class DerivativesDataSink(bid_derivative):
```
mattcieslak, line 39: This class keeps on getting redefined in various places. Note to self: we should remove these redefinitions

xcp_d/workflow/anatomical.py
```diff
@@ -1,784 +1,1371 @@
...
+        )
+
+    else:
+        all_files = list(layout.get_files())
+
+        t1w_transform_wf = pe.Node(
+            ApplyTransformsx(
+                num_threads=2,
+                reference_image=mnitemplate,
+                # transforms=[str(t1w_to_mni), str(MNI92FSL)], #TM: need to replace MNI92FSL xfm with the correct xfm from the MNI output space of fMRIPrep/NiBabies (MNI2009, MNIInfant, or for cifti output MNI152NLin6Asym) to MNI152NLin6Asym.
```
mattcieslak, line 240: @madisoth did you add this comment or was it in there before? Should we open a PR for this?

xcp_d/workflow/anatomical.py
```diff
@@ -1,784 +1,1371 @@
...
+            )
+
+            select_inv_z_component = pe.Node(
+                niu.Select(index=[2]),
+                name="select_inv_z_component",
+                mem_gb=mem_gb,
+                n_procs=omp_nthreads,
+            )
+
+            # reverse y-component of the warpfield
```
mattcieslak, line 523: do you have a reference for this? It sounds right but would be good to have a link to someone confirming it

mattcieslak (commented):
Only minor comments, looks great!

xcp_d/interfaces/c3.py
madisoth: The "multicomp_split" trait in Nipype's c3d interface is incorrect (at least for the version of c3d I put in xcpd_build), so I made this modified copy of the interface for now

PR #383, opened 2022-07-15
Replace NiPype's filemanip.py

madisoth:

Replaces NiPype's filemanip.py with a modified version, which adds CIFTI file extensions to the built-in list of "special extensions". 

"Special extensions" are multi-part extensions (e.g. ".nii.gz", ".dtseries.nii") which NiPype is directed to treat as a single extension for filemanip functions.

This fixes an issue in 0.1.0 where using --cifti without --despike would error, due to the censored BOLD from CensorScrub having a bad file extension of ".dtseries_bold.nii". Also fixes other cases of CIFTI files in the work directory with bad extensions. 

Also corrected a misindented try-except-else in filemanip.py (L746)


xcp_d/interfaces/connectivity.py
```diff
@@ -112,28 +120,31 @@ def get_atlas_nifti(atlasname):
...
+    elif atlasname == "tiansubcortical":
         atlasfile = pkgrf(
-            'xcp_d',
-            'data//niftiatlas/TianSubcortical/Tian_Subcortex_S3_3T.nii.gz')
+            "xcp_d", "data//niftiatlas/TianSubcortical/Tian_Subcortex_S3_3T.nii.gz"
+        )
     else:
-        raise RuntimeError('atlas not available')
+        raise RuntimeError("atlas not available")
     return atlasfile
```
mattcieslak, line 148: Note to self: this should get a refactor to be more readable

xcp_d/interfaces/prepostcleaning.py
```diff
@@ -67,48 +79,52 @@ def _run_interface(self, runtime):
         # Check if we need to do anything
         if self.inputs.initial_volumes_to_drop == 0:
             # write the output out
-            self._results['bold_file_dropped_TR'] = self.inputs.bold_file
-            self._results['fmriprep_confounds_file_dropped'
-                          '_TR'] = self.inputs.fmriprep_confounds_file
+            self._results["bold_file_dropped_TR"] = self.inputs.bold_file
+            self._results[
+                "fmriprep_confounds_file_dropped" "_TR"
```
mattcieslak, line 84: This is an odd way to do it. Did you also run a ton of auto-linting for this PR?

xcp_d/utils/filemanip.py
```diff
@@ -0,0 +1,985 @@
...
+        ".pscalar.nii",
+        ".ptseries.nii",
+        ".sdseries.nii",
+        ".label.gii",
+        ".label.gii",
+        ".func.gii",
+        ".shape.gii",
+        ".rgba.gii",
+        ".surf.gii",
+    ]
```
mattcieslak, line 114: nice, this would be worth doing a PR on nipype at some point too

xcp_d/utils/filemanip.py
```diff
@@ -0,0 +1,985 @@
...
+
+
+@contextlib.contextmanager
+def indirectory(path):
+    cwd = os.getcwd()
+    os.chdir(str(path))
+    try:
+        yield
+    finally:
+        os.chdir(cwd)
```
mattcieslak, line 985: do we need all of the other functions in here? Or just the one that supports the cifti extensions?

mattcieslak (commented):
In the future could you do separate PRs for linting-related changes and workflow-related changes? I assumed that only filemanip.py and files that now import from it are the ones with content changes

madisoth:
@mattcieslak agree that this PR should be cleaned up a bit -- I'll convert this PR to a draft for now and resubmit later

mattcieslak (approved):
Excellent!!

PR #388, opened 2022-07-27
ENH: Update filemanip.py "special_extensions" with the rest of the CIFTI-2 multi-part extensions

madisoth:
adds the rest of the CIFTI-2 extensions to special_extensions

- filemanip.py: add the rest of the multi-part CIFTI-2 extensions to the list "special_extensions" so that filemanip functions treat them like single-part extensions.


mattcieslak (approved):
Looks good, thanks for adding these

PR #442, opened 2022-08-29
Fix fsLR32k reg and ApplyWarpfield issues

madisoth:

Fixes misregistration of anat surfs to fsLR32k. Also fixes surface_apply_warpfield failures due to surfaces exceeding the warp FOV.

- add SurfaceSphereProjectUnproject step (concatenate sphere xfms from native -> 164k -> 32k)
- add default outfile name to SurfaceSphereProjectUnproject interface
- input native (not MNI-aligned) surfaces to fsLR32k CiftiSurfaceResample
- revert T1w-MNI surface alignment xfm to two steps (affine 1st / displacement field 2nd) instead of first combining warps -- prevents xfm failure due to surfaces extending outside the warp FOV when the affine component of the warp is severe and/or the original T1w's brain is too close to edge of FOV
- streamlined some redundant per-hemi / per-surface code with MapNode / iterables


xcp_d/workflow/anatomical.py
```diff
@@ -28,11 +27,15 @@
 from nipype.interfaces.fsl.maths import BinaryMaths as fslbinarymaths  # TM
 from nipype.interfaces.fsl import Merge as fslmerge  # TM
 from ..interfaces.c3 import C3d  # TM
-from ..interfaces.ants import CompositeInvTransformUtil  # TM
+from ..interfaces.ants import CompositeInvTransformUtil, ConvertTransformFile  # TM
```
mattcieslak, line 30: No need to add initials to changed lines, git will take care of this

xcp_d/workflow/anatomical.py
```diff
@@ -573,84 +596,125 @@ def init_anatomical_wf(
...
             )
 
-            join_lh_32k_surfs = pe.JoinNode(
-                niu.Merge(1),
-                name="join_lh_32k_surfs",
-                joinsource="lh_surface_apply_warpfield",
-                joinfield="in1",
+            # collect the fsLR32k mid, pial, wm surfs per hemi
+
+            lh_32k_surf_wf = pe.MapNode(
```
mattcieslak, line 695: In general if there are only a few things to iterate over, it's better to make explicitly-named nodes instead of using a mapnode. It's more verbose but easier to track down errors and easier to connect their outputs to datasinks

mattcieslak (commented):
Only minor cosmetic changes here, but this looks great.

Maybe not as part of this PR, but definitely soon we should add a pytest that checks for the specific files that this workflow produces

PR #458, opened 2022-09-02
Update utils.py to fix MNILin6 imports

smeisler:
Fixing MNI import for Lin6 (typo in "Asym")


mattcieslak:
Good catch @smeisler thank you!!

PR #469, opened 2022-09-09
[DOC] Address formatting issues in documentation

tsalo:
This doesn't address all formatting issues, but it does tackle a few.

- Outside of the repo, I set the RTD project to build pull requests, so it should show up in the status checks from now on.
- Remove an unnecessary requirements file.
- Remove an extra RTD config file.
- Move the pull request template into the `.github` folder.
- Use Napoleon to match fMRIPrep's formatting, and to allow custom docstring sections (e.g., Inputs, Outputs).
- Address miscellaneous formatting issues in the docs folder. I haven't gotten all of them, and there are also some docstrings that need to be updated, but I think I'll tackle those in a separate PR.

A brief look-through of everything would be helpful.


tsalo:
It looks like the workflow graphs aren't rendering, but AFAICT they don't render in `latest` either.
Here's what it looks like in the current PR:

image

Part of the problem seems to be that the workflow has required arguments that are not specified in the graph call. The other issue is that the workflow will then run to generate the graph. I think we need to (1) fill out the workflow graph calls with actual values and (2) adopt something like fMRIPrep's [`mock_config `](https://github.com/nipreps/fmriprep/blob/91dae68908360ce8ab89d6e43f16a62fd6deb1d2/fmriprep/workflows/tests.py#L33) to handle it all safely.

I think both things can be addressed in a separate PR though.

kahinimehta:
@tsalo Yes, I think that might be worth tackling separately - I just made an issue for it

PR #482, opened 2022-09-17
[MAINT] Add CITATION.cff file

tsalo:
Closes None.

- Add a CITATION.cff file with information from the Zenodo page.
- Add a documentation page describing the release process.

The new developers guide page.


CITATION.cff
```diff
@@ -0,0 +1,77 @@
...
+    family-names: Sydnor
+    affiliation: University of Pennsylvania
+    orcid: 'https://orcid.org/0000-0002-8640-668X'
+  - given-names: Sydney
+    family-names: Covitz
+    affiliation: University of Pennsylvania
+    orcid: 'https://orcid.org/0000-0002-7430-4125'
+identifiers:
+  - type: doi
+    value: 10.5281/zenodo.5139942
```
tsalo, line 54: The one thing I don't have a solution for is updating the Zenodo DOI. The version-specific DOI is created when the GitHub release is published, but we would need to update the Zenodo DOI in the CITATION.cff file before the release if we want folks to cite version-specific DOIs...

CITATION.cff
```diff
@@ -0,0 +1,77 @@
...
+    affiliation: University of Minnesota
+    orcid: 'https://orcid.org/0000-0003-3030-6580'
+  - given-names: Valerie
+    family-names: Sydnor
+    affiliation: University of Pennsylvania
+    orcid: 'https://orcid.org/0000-0002-8640-668X'
+  - given-names: Sydney
+    family-names: Covitz
+    affiliation: University of Pennsylvania
+    orcid: 'https://orcid.org/0000-0002-7430-4125'
```
tsalo, line 51: Should Ted and/or Damien be added to the end?

kahinimehta: I believe so? @mattcieslak

kahinimehta (approved):

CITATION.cff
tsalo: Sounds good. Does it look like anyone else is missing?

tsalo: Per today's DCAN call, this is something we can bring up in Friday's call.

tsalo: We didn't get around to it, so I'm going to merge this and we can update it in the future if necessary.

PR #493, opened 2022-09-21
[RF] Get rid of unused variables

kahinimehta:
- Remove variables that are not being used

### DIFF
--- a/xcp_d/interfaces/layout_builder.py
+++ b/xcp_d/interfaces/layout_builder.py
@@ -172,7 +172,6 @@ class Section(object):
     img_path : str, optional
     regs_slider : None or str, optional
     img_modal : None or str, optional
-    kwargs : dict, optional
     """
 
     def __init__(self,
--- a/xcp_d/interfaces/layout_builder.py
+++ b/xcp_d/interfaces/layout_builder.py
@@ -254,10 +254,10 @@ class TasksSection(Section):
 
         self.run(tasks)
 
-    def write_t1_reg_rows(self, task_name):
+    def write_t1_reg_rows(self, task_name, task_num):
         """Write T1w rows."""
         # Write the header for the next few rows.
-        self.section += constants.TASK_LABEL_ROW.format(task_name=task_name)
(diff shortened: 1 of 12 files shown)


xcp_d/workflow/restingstate.py
```diff
@@ -167,7 +167,6 @@ def init_compute_alff_wf(mem_gb,
 
 def init_surface_reho_wf(mem_gb,
                          omp_nthreads,
-                         smoothing,
```
tsalo, line 170: Can you also remove smoothing from the parameter list in the docstring?

xcp_d/workflow/postprocessing.py
```diff
@@ -24,18 +24,12 @@
...
         upper_bpf,
         bpf_order,
         smoothing,
         bold_file,
         params,
-        motion_filter_type,
-        motion_filter_order,
-        band_stop_max,
-        band_stop_min,
-        initial_volumes_to_drop,
```
tsalo, line 38: Some of these need to be removed from the example call and the parameters list in the docstring.

xcp_d/interfaces/layout_builder.py
```diff
@@ -256,11 +255,10 @@ def __init__(self, tasks=[], img_path='./figures', **kwargs):
 
         self.run(tasks)
 
-    def write_t1_reg_rows(self, task_name, task_num):
+    def write_t1_reg_rows(self, task_name):
         """Write T1w rows."""
         # Write the header for the next few rows.
-        self.section += constants.TASK_LABEL_ROW.format(task_name=task_name,
-                                                        task_num=task_num)
+        self.section += constants.TASK_LABEL_ROW.format(task_name=task_name)
```
tsalo, line 261: Why remove `task_num` from this method?

tsalo (changes requested):
A few minor things. I think in some cases you need to remove the arguments from the docstrings as well.

xcp_d/interfaces/layout_builder.py
kahinimehta: I don't think it was being used

tsalo (approved):
Happy to merge this once CI passes.

PR #499, opened 2022-09-21
[REF] Empty __init__.py files

tsalo:
Related to #498. This drops the number of Sphinx warnings from 80 to 58, because `apidoc` tries to render duplicate entries (e.g., `interfaces.bids.X` and `interfaces.X`).

- Remove submodule imports from __init__ files. My understanding is that including these imports in the init file is useful for making them accessible without the need for fully-specified imports (i.e., all the way down to the file). However, almost nothing in `xcp_d` is meant to be accessed by users, so streamlining their imports isn't necessary. We also have to deal with circular import issues when we deal with aliased imports.
- Fix any lingering import issues.

The API docs.


tsalo:
@kahinimehta @madisoth do either of you have any issue with me merging this?

kahinimehta (approved):

PR #521, opened 2022-09-26
[TST] Store CircleCI artifacts

tsalo:
Closes None, but addresses something discussed offline.

- Store the artifacts (just derivatives) for each of the CircleCI tests.


tsalo:
Merging this without a review since it's a very small change, and I've checked the artifacts.

Issue #338, opened 2022-06-11
[ENH] Docker image is too big


mattcieslak:
The docker image is huge, I don't think we need all of fsl or so many copies of the FSL atlases

tsalo:
We've removed FSL, but it's still really big. We should review the contents of the image again to see why it's still so big.

tsalo:
Closing in favor of https://github.com/PennLINC/xcpd_build/issues/21.

PR #528, opened 2022-09-26
[REF] Remove FSL from dependencies

### LINKED ISSUE #524: Consider dropping FSL dependency
## Summary

Apparently FSL is pretty big, and we only seem to be using it for smoothing and file concatenation. If we're not leveraging any of FSL's fancier tools (e.g., SUSAN), then we might as well drop the dependency completely.

tsalo:
Closes #524. Related to PennLINC/xcpd_build#2. Also related to #338.

- Replace calls to fslmerge with nilearn.image.concat_imgs.
- Replace calls to fslsmooth with nilearn.image.smooth_img.
- Replace calls to fslmaths with nilearn.image.math_img.

None.


kahinimehta:
@tsalo might need to update the smoothing tests to reflect the removal of FSL

tsalo:
@kahinimehta done! Just waiting for the Docker image to deploy and then I'll update the Dockerfile.

kahinimehta (approved):

PR #530, opened 2022-09-26
Improve confound plot readability and fix moving registration plot

kahinimehta:
Attempting to fix issues with the Executive Summary

Changes proposed:

- Modify parameters of confounds plot to improve readability.
- Fix shifting registration plot that is currently only included in the executive summary. This plot was only showing the moving SVG, but not the target SVG, before.

### DIFF
--- a/xcp_d/data/reports.yml
+++ b/xcp_d/data/reports.yml
@@ -16,41 +16,41 @@ sections:
 - name: About
   reportlets:
   - bids: {datatype: figures, desc: about, suffix: bold}
-- name: Executive Summary
-  ordering: session,task,acquisition,ceagent,reconstruction,direction,run,echo
-  title: <sub>0</sub>_<task>0</task>_<space>0</space>_<run>0</run>
-  reportlets:
-  - bids:
-      datatype: figures
-      desc: brainplot
-      extension: [.html]
-      suffix: T1W
-      title: Anatomical Data
-      caption: T1W
-      static: false
-  - bids:
-      datatype: figures
-      desc: brainplot
-      extension: [.html]
-      suffix: T2W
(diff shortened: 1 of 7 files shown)


xcp_d/interfaces/layout_builder.py
```diff
@@ -3,7 +3,7 @@
 
 __doc__ = """
 Builds the layout for the Executive Summary of the bids-formatted output from
-the DCAN-Labs fMRI pipelines.
+the DCAN-Labs fMRI pipelines."
```
tsalo, line 6: Is this a typo?

xcp_d/data/reports.yml
```diff
@@ -16,3 +16,41 @@ sections:
 - name: About
   reportlets:
   - bids: {datatype: figures, desc: about, suffix: bold}
+# - name: Executive Summary
```
tsalo, line 19: Do you want to keep these changes in?

tsalo (commented):
I think the changes look good, but are reduced from what the PR comment says (i.e., using NiPreps config), so could you update that summary?

xcp_d/data/reports.yml
kahinimehta: I think so, in case we want to build on them later - what do you think?

xcp_d/interfaces/layout_builder.py
kahinimehta: yes, will change

tsalo (approved):
LGTM!

xcp_d/data/reports.yml
tsalo: Fair enough. Happy to keep them in.

PR #542, opened 2022-09-29
[TST] Check for existence of outputs in tests

tsalo:

- Add a step to the CircleCI workflow tests that compares the outputs from the tests to a pre-generated list of outputs. This will flag any changes that produce different outputs.
- Add lists of outputs for each of the workflows.

None.


tsalo:
It's passing!

tsalo:
Since this is just an additional test, and it wouldn't pass if there was a problem, I'm going to merge it without reviews.

PR #545, opened 2022-09-30
[TST] Make local tests stricter

tsalo:

- Ensure that the local xcp_d path exists, and raise an error if it doesn't. Previous it would just raise a warning, then continue on using the version in the Docker file.
- Don't call `setopt` unless the shell is zsh, instead of the opposite.
- Make sure a few other things in `get_data.sh` are treated as strings.

None.

### DIFF
--- a/.circleci/get_data.sh
+++ b/.circleci/get_data.sh
@@ -119,7 +119,7 @@ The results of running a colornest subject through fmriprep
 freesurfer_colornest:
 ---------------------
 
-The freesurfer results for the same data as in `fmriprep_colornest`
+The freesurfer results for the same data as in "fmriprep_colornest"
 
 fsaverage*:
 -----------
--- a/.circleci/get_data.sh
+++ b/.circleci/get_data.sh
@@ -1,5 +1,5 @@
 
-if [[ "$SHELL" =~ zsh ]]; then
+if [[ "$SHELL" == zsh ]]; then
   setopt SH_WORD_SPLIT
 fi
 
--- a/.circleci/get_data.sh
+++ b/.circleci/get_data.sh
@@ -5,22 +5,32 @@ fi
 
 # Edit these for project-wide testing
 WGET="wget --retry-connrefused --waitretry=5 --read-timeout=20 --timeout=15 -t 0 -q"
-LOCAL_PATCH= ~/projects/xcp_d/xcp_d  # Change to your local patch 
-IMAGE=pennlinc/xcp_d:unstable
+LOCAL_PATCH="~/projects/xcp_d/xcp_d"  # Change to your local patch
+IMAGE="pennlinc/xcp_d:unstable"
 
 # Determine if we're in a CI test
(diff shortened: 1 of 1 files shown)


.circleci/get_data.sh
```diff
@@ -1,26 +1,36 @@
 
-if [[ "$SHELL" =~ zsh ]]; then
+if [[ "$SHELL" == zsh ]]; then
   setopt SH_WORD_SPLIT
 fi
```
tsalo, line 4: I think `setopt` is a zsh-specific command.

.circleci/get_data.sh
```diff
@@ -1,26 +1,36 @@
...
 
 # Edit these for project-wide testing
 WGET="wget --retry-connrefused --waitretry=5 --read-timeout=20 --timeout=15 -t 0 -q"
-LOCAL_PATCH= ~/projects/xcp_d/xcp_d  # Change to your local patch 
-IMAGE=pennlinc/xcp_d:unstable
+LOCAL_PATCH="~/projects/xcp_d/xcp_d"  # Change to your local patch
```
tsalo, line 8: I noticed that the space was causing problems sometimes. Valid paths weren't being found.

.circleci/get_data.sh
```diff
@@ -1,26 +1,36 @@
...
   IN_CI="false"
   NTHREADS=2
   OMP_NTHREADS=2
+
+  # check that the local xcp_d path exists
+  if [ ! -d $LOCAL_PATCH ]
+  then
+    echo "Path $LOCAL_PATCH DNE"
+    exit 125
+  fi
```
tsalo, line 32: This is the key change. I accidentally kept running tests locally with the wrong local path. The tests raised a warning about the directory not existing, but I missed it because they continued running using the version of `xcp_d` in the Docker image.

.circleci/get_data.sh
```diff
@@ -109,7 +119,7 @@ The results of running a colornest subject through fmriprep
 freesurfer_colornest:
 ---------------------
 
-The freesurfer results for the same data as in `fmriprep_colornest`
+The freesurfer results for the same data as in "fmriprep_colornest"
```
tsalo, line 122: This was considered a command instead of a string because of the backticks.

.circleci/get_data.sh
mattcieslak: it is - this only runs if the shell is zsh though

tsalo: Really? Does `!=` mean that it only runs if it _isn't_ zsh? I keep getting `./get_data.sh: line 3: setopt: command not found` when running with bash.

mattcieslak: AH I misread the text above. Yes we only want this if it's being run in zsh!

.circleci/get_data.sh
mattcieslak: great catch

mattcieslak (approved):
thank you!!

PR #562, opened 2022-10-09
Replace `--func-only` with `--warp-surfaces-native2std`

tsalo:
Closes None.

- Invert the logic of `--func-only` and replace it with `--warp-surfaces-native2std`. Now, the surfaces won't be outputted by default, unless you're running `hcp` or `dcan` data.
- Split the steps that warp the t1w and t1seg files to MNI space out of the anatomical workflow and into a separate one (`init_t1w_wf`) that runs no matter what.
- Move the creation of the T1 brain sprite figure into a separate workflow as well (`init_brainsprite_wf`).
- So much refactoring!

### Breaking changes

- `space-MNI152NLin6Asym_desc-preproc_dseg.nii.gz` --> `space-MNI152NLin6Asym_dseg.nii.gz`
    - The dseg output no longer takes its name from the T1w file. It now uses its own original file as the name source. This means that fields unique to the T1w file (e.g., `desc-preproc` for fMRIPrep derivs) will go away.
- The anatomical workflow is no longer run by default.


xcp_d/utils/doc.py
```diff
@@ -209,6 +209,34 @@
...
+        * - ``<source_entities>_space-fsLR_den-32k_hemi-<L|R>_hcpmidthickness.surf.gii``
+          - Something
+        * - ``<source_entities>_space-fsLR_den-32k_hemi-<L|R>_hcpveryinflated.surf.gii``
+          - Something
+        * - ``<source_entities>_space-fsLR_den-32k_hemi-<L|R>_midthickness.surf.gii``
+          - Something
+        * - ``<source_entities>_space-fsLR_den-32k_hemi-<L|R>_pial.surf.gii``
+          - Something
+        * - ``<source_entities>_space-fsLR_den-32k_hemi-<L|R>_smoothwm.surf.gii``
+          - Something
```
tsalo, line 237: @madisoth could you fill in the gaps on this? I don't know how to describe each, or what they're used for.

xcp_d/utils/doc.py
```diff
@@ -209,6 +209,34 @@
     List of paths to atlas-specific ROI-to-ROI correlation files.
 """
 
+docdict["process_surfaces"] = """
+process_surfaces : :obj:`bool`, optional
+    If True, a workflow will be run to warp native-space (fsnative) reconstructed cortical
+    surfaces (surf.gii files) produced by Freesurfer into standard (fsLR) space.
+    These surface files are primarily used for visual quality control.
```
tsalo, line 216: @madisoth are they used for anything other than visual QC?

tsalo:
@kahinimehta I've made the change you requested (selecting the brainsprite ribbon outside of the anatomical workflow), and tests are finally passing. When you have a chance, could you take another look?

kahinimehta (commented):
Just confirming the right file is being used for the "ribbon", if the anat workflow is not run - as we spoke about yesterday

xcp_d/workflow/base.py
```diff
@@ -97,7 +99,7 @@ def init_xcpd_wf(
                 work_dir=".",
                 dummytime=0,
                 fd_thresh=0.2,
-                func_only=False,
+                process_surfaces=True,
```
kahinimehta, line 102: isn't false the default?

kahinimehta (approved):

xcp_d/workflow/base.py
tsalo: Good point! I'll switch it to the default.

PR #570, opened 2022-10-12
Parallelize main workflow

tsalo:


tsalo:
Not sure this is worth doing. The biggest problem is that the boilerplate depends on info within the files, so we need them as variables in the workflow, which kind of defeats the purpose of making them connections instead.

PR #575, opened 2022-10-13
Fix documentation and validation of motion filtering parameters

### LINKED ISSUE #569: [BUG] Add an error message if motion filter parameters are not in the right range
See https://neurostars.org/t/question-re-filtering-fd-values/23751

tsalo:
Closes #569.

- Raise a warning if motion filter min/max values are suspiciously low (<1).
- Raise an error if motion filter min value is higher than max value.
- Raise a warning if either min or max motion filter value is defined when the motion filter itself is disabled.
- Raise a warning if max motion filter value is defined when the motion filter is set to "lp", since frequencies below min should be retained.
- Set the default for min/max motion filter values to None instead of 0.
- Fix documentation of motion filter values.
- Raise an error if min _bandpass filter_ value is higher than max value.
- Mention the filtering done in the denoising workflows' boilerplates.
- Don't mention mean-centering and detrending nuisance regressors in denoising workflows' boilerplates, because the nuisance regressors are _not_ mean-centered or detrended.
- In CensorScrub interface, replace low_freq and high_freq inputs with band_stop_min and band_stop_max.
- In QCPlot interface, replace low_freq and high_freq inputs with band_stop_min and band_stop_max.

- The general workflow page
- The workflow API pages


xcp_d/workflow/cifti.py
```diff
@@ -147,46 +147,68 @@ def init_ciftipostprocess_wf(
...
-Both the nuisance regressors and volumes were demean and detrended. Furthermore, any volumes
-with framewise-displacement greater than {fd_thresh} mm [@power_fd_dvars;@satterthwaite_2013] were
-flagged as outliers and excluded from nuisance regression.
-"""
-
-    else:
-        workflow.__desc__ = workflow.__desc__ + f""" \
-before nuisance regression and filtering,both the nuisance regressors and volumes were demeaned
-and detrended. Volumes with framewise-displacement greater than {fd_thresh} mm
+        initial_volumes_to_drop = int(np.ceil(dummytime / TR))
```
tsalo, line 196: This is changed from `int(np.floor(dummytime / TR))` to match the nifti workflow's approach. Does anyone know if there's a reason the cifti workflow uses `numpy.floor`, while the nifti workflow uses `numpy.ceil`?

kahinimehta: definitely a typo - they should both be ceil

kahinimehta (approved):

tsalo:
Thanks @kahinimehta! I'll merge once CI finishes up.

PR #579, opened 2022-10-14
Add censoring plot to summary reports

### LINKED ISSUE #511: Use stricter FD threshold in CI test
## Summary

Use a stricter FD threshold (e.g. 0.1) in the current CI tests, or add an additional test(s) with a stricter threshold

## Additional details

A stricter threshold ensures that censoring and interpolation are tested. The colornest subject has no frames with FD above the default 0.2 threshold (assuming radius=40 as used in the current tests), so no censoring or interpolation is applied by the current tests. 

Also since regression is currently applied after censoring but before interpolation, this ensures we have a test case where the post-regression BOLD is a different length than the pre-reg/resid BOLD for catching potential issues (e.g. with dvars plotting)

### LINKED ISSUE #576: QCPlot shows FD after filtering
## Summary

Based on the code below, it looks like QCPlot uses the filtered framewise displacement values in both the preprocessing and postprocessing carpet+line plots. This is probably misleading to users, since it seems like no filtering was done between the two versions.

https://github.com/PennLINC/xcp_d/blob/d4f8bd01fa0cbeaaf23420f7b142a603469440d4/xcp_d/interfaces/qc_plot.py#L104-L119

tsalo:
Closes #511. Related to #576. Namely, while this doesn't address the problem with the QCPlot (I've shifted those changes to #581), it does add a new figure to the report that should make it clearer what the filtering was used for.

- Make `motion_filter_type`, `band_stop_min`, and `band_stop_max` optional parameters in `xcp_d.utils.confounds.load_motion`. The defaults are all None.
- Apply the notch filter in the NiftWithFreeSurfer test.
- Create a new `CensoringPlot` interface, which generates a line plot showing the unfiltered FD, filtered FD, any censored volumes, any dummy volumes, and the outlier threshold.
- Add the new CensoringPlot to the post-processing workflows and summary reports.
- Mention interpolation and despiking in boilerplates.
    - The despike mention is too brief.
    - The interpolation mention is also too brief, and honestly I'm having trouble making sense of the whole interpolation procedure.


madisoth:
I actually had implemented this (unfiltered vs. filtered FD in the pre- and post-plots) in a dev branch a few months ago, but got feedback from Damien that he wanted filtered FDs on **both** the pre- and post- plots.  IIRC part of the rationale was that the FD filtering isn't part of the BOLD regression workflow, but this would be good to discuss at our biweekly meeting to make sure everyone's on the same page.

tsalo:
@madisoth that's a good point. I want to make a release early next week and this filtering thing _is_ confusing users, but implying that the filtered FD and/or motion parameters were used for denoising would confuse folks even more. Instead, I'll mock up a censoring figure with the unfiltered and filtered FD time series and vertical bands at all of the flagged volumes.

tsalo:
This is just a first pass, but something like this:

image

xcp_d/utils/plot.py
```diff
@@ -512,9 +512,12 @@ def plot_svgx(rawdata,
 
     # Remove first N deleted from raw_data so it's same length as censored files
     if len(raw_dvars_data) > len(residual_dvars_data):
+        # TODO: Should this be [-len(residual_dvars_data):] ?
+        # ... Seems to grab first N, not last N.
         raw_dvars_data = raw_dvars_data[0:len(residual_dvars_data)]
         raw_data = raw_data[:, 0:len(residual_dvars_data)]
-        # regressed_dvars_data = raw_dvars_data #TODO: Check if this is needed
+        # TODO: Figure out how to slice the regressed DVARS instead of just overwriting it
+        regressed_dvars_data = raw_dvars_data
```
tsalo, line 520: @kahinimehta it looks like you added the comment in #447. Following up here, since the mismatch in length between `raw_dvars_data` and `regressed_dvars_data` was raising an error in this PR a few lines down where the DataFrame is created.

I don't know the full set of conditions that cause a problem, but at least one of them is having >0 volumes censored. I think the way the cropping is done here is wrong, and I have added a TODO comment to that effect. I also think it's weird to overwrite the regressed DVARS with the raw DVARS, but I'm happy to keep the hack in for now, since it's outside the scope of this PR.

xcp_d/interfaces/qc_plot.py
```diff
@@ -22,7 +23,175 @@
...
+                time_array,
+                filtered_fd_timeseries,
+                label="Filtered Framewise Displacement",
+                color="red",
+            )
+        else:
+            filtered_fd_timeseries = preproc_fd_timeseries.copy()
+
+        # NOTE: TS- Probably should replace with the actual tmask file.
+        tmask = filtered_fd_timeseries >= self.inputs.fd_thresh
```
kahinimehta, line 153: Yes, would agree - not sure if we wanna do that in this PR/ low priority?

xcp_d/utils/modified_data.py
```diff
@@ -65,7 +65,7 @@ def interpolate_masked_data(bold_data, tmask, TR=1):
     flagged by the temporal mask.
 
     NOTE: TS- Why are slice times being inferred from the number of volumes?
-    Am I missing somehting?
+    Am I missing something?
```
kahinimehta, line 68: I think it's just bad commenting/naming - if there's something that makes more sense, we can definitely change it

xcp_d/utils/plot.py
```diff
@@ -512,9 +512,12 @@ def plot_svgx(rawdata,
 
     # Remove first N deleted from raw_data so it's same length as censored files
     if len(raw_dvars_data) > len(residual_dvars_data):
+        # TODO: Should this be [-len(residual_dvars_data):] ?
+        # ... Seems to grab first N, not last N.
```
kahinimehta, line 516: Yes, seems like this is the better way to implement it

xcp_d/workflow/anatomical.py
```diff
@@ -1477,5 +1477,18 @@ def init_anatomical_wf(
                 "No FreeSurfer derivatives detected. "
                 "Surface transformation will not be performed."
             )
+            # The inputnode needs to be connected to *something* to be added to the workflow.
+            # This "nothingnode" exists just to allow the inputnode to connect to something.
```
kahinimehta, line 1481: That would make sense

xcp_d/utils/modified_data.py
tsalo: Hopefully we'll be able to just drop this whole step soon in favor of Nilearn.

xcp_d/interfaces/qc_plot.py
tsalo: It's pretty low priority given how simple the censoring is. I'd rather tackle it in a future PR.

tsalo:
@kahinimehta are there any aesthetic changes you'd like to see in the censoring plot?

xcp_d/utils/plot.py
tsalo: I've punted the issue to #589. Hopefully we can find a solution, but I don't want to address it in this PR.

xcp_d/utils/plot.py
tsalo: I've punted the issue to https://github.com/PennLINC/xcp_d/issues/589. This part of the bug is easy enough to fix, but I'd like to address it in a separate PR.

kahinimehta (approved):

PR #581, opened 2022-10-17
Show framewise displacement before filtering in QCPlot

### LINKED ISSUE #576: QCPlot shows FD after filtering

tsalo:
Closes #576.

- Show the unfiltered framewise displacement time series in the pre- and post-processing QC plots. This is the most important change.
- Refactor QCPlot to be much simpler.
- Remove unused inputs to QCPlot. The interface doesn't need info about the motion filter.
- Add comments in places where I'm confused.


kahinimehta (approved):

tsalo:
Thanks @kahinimehta!

PR #522, opened 2022-09-26
Enh/concat_new

madisoth:
## Changes proposed in this pull request

This PR replaces the unmerged #417. Addresses parts of #311 and #450 plus other issues with the per-task concatenation option (`--combineruns / -m`)

- Fix for some cases where concatenation with `--combineruns / -m` could fail when a file's BIDS task label (the "XXXX" in `_task-XXXX`) is a substring of another part of the file's path/name
- Fix error concatenating runs per-task when a session has multiple BIDS tasks. 
- Proper handling of tasks with a single run when `--combineruns / -m` is used
- Output the `-DCAN.hdf5` motion data file for both the concatenated data and each individual run when `--combineruns / -m` is used, instead of for concatenated data only as currently done.
- ~~FD censoring mask from CensorScrub is now output with the functional derivatives, as `*-desc-tmask_bold.tsv`~~ Moved to #586 and renamed to `*_outliers.tsv`, per [BEP012](https://bids-specification--519.org.readthedocs.build/en/519/05-derivatives/05-functional-derivatives.html#temporal-outlier-masks).
- Revise FD plot in executive summary (increased height of FD plot frame, reduced size of marker dots, add y-axis ticks/labels at 0.1 mm intervals, match the color of the frames-under-FD-threshold text to the marker dots)

PR #586, opened 2022-10-18
Output temporal mask

tsalo:
Closes None, but breaks off a part of #522 to reduce scope.

- Write out the temporal mask as `<source_entities>_outliers.tsv`.


tsalo:
The new outliers.tsv file (and motion.tsv file, for that matter) doesn't include a header. I think I should tackle that in a separate PR though.

kahinimehta (approved):

PR #588, opened 2022-10-18
Deprecate `--bandpass_filter` in favor of `--disable-bandpass-filter`

### LINKED ISSUE #577: Deprecate `--bandpass_filter` in favor of `--disable-bandpass-filter`
## Summary

The `--bandpass_filter` parameter accepts a boolean and defaults to True, so users can only disable it with `--bandpass_filter False`. It would be much simpler to just have a flag with no associated argument, like `--disable-bandpass-filter`.

tsalo:
Closes #577.

- Change current DeprecationWarnings from 0.1.6 to 0.3.0.
- Deprecate the `--bandpass_filter` CLI parameter in favor of the simpler `--disable-bandpass-filter` flag.

None


xcp_d/cli/run.py
```diff
@@ -255,13 +255,28 @@ def get_parser():
     )
 
     g_filter = parser.add_argument_group("Filtering parameters and default value")
-    g_filter.add_argument(
+
+    bandpass_filter_params = g_filter.add_mutually_exclusive_group()
+    bandpass_filter_params.add_argument(
```
tsalo, line 260: This parameter has to go first to set the default, I think. It looks like it's "first come, first served" when it comes to defining defaults on the same destination variable. Not really a problem, just a funny idiosyncrasy to keep in mind.

kahinimehta (approved):

tsalo:
Thanks! Merging now.

Issue #578, opened 2022-10-14
Only one brainsprite figure supported in current report


tsalo:
## Summary

After some local testing, it looks like the main HTML reports can only handle one brainsprite figure at a time. If I request the ALFF figure, then the ReHo one, only the first run's ALFF figure will show up. If I request the ReHo one first, then only the first run's ReHo figure will render.

This is probably a bug on niworkflows' side, since the Report class is ultimately imported from there. However, it could be something that was resolved in a more recent niworkflows version, or they might even be unaware of the problem.

tsalo:
Look at [`nipreps/dmriprep-viewer`](https://github.com/nipreps/dmriprep-viewer), which includes multiple brainsprites in its reports.

tsalo:
We've dropped brainsprite from the report, so I'm going to close this for now.

Issue #519, opened 2022-09-26
[ENH] General report 


kahinimehta:
- [ ] Potentially allow for toggling between the before/after .svg plots

kahinimehta:
@tsalo will have a look soon

PR #590, opened 2022-10-18
Add correlation heatmaps plot to HTML report

tsalo:
Related to #519 and other reporting issues. 

I noticed that the bbregister plot that exists in the executive summary is just copied from fMRIPrep. Do we want to include it in the xcpd report? Users could just refer back to their fMRIPrep derivatives, and anything they'd catch from the bbregister plot should probably have been caught before running xcpd, IMHO.

I would love to add the ReHo and ALFF plots to the report, but #578 still holds.

- Add the `connectivityplot` to the HTML report.

### DIFF
--- a/xcp_d/data/reports.yml
+++ b/xcp_d/data/reports.yml
@@ -4,15 +4,21 @@ sections:
   ordering: session,task,acquisition,ceagent,reconstruction,direction,run,echo
   reportlets:
   - bids: {datatype: figures, desc: summary, suffix: bold}
-  - bids: {datatype: figures, desc: qualitycontrol,suffix: bold}
-  - bids: {datatype: figures, desc: preprocessing,suffix: bold}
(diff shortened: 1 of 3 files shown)


kahinimehta (approved):

tsalo:
From Ted:

- Drop boldref figure.
- Add T1 brainsprite (should be fine as long as we don't try to add any other brainsprites).

xcp_d/interfaces/qc_plot.py
```diff
@@ -389,7 +389,7 @@ def _run_interface(self, runtime):
             )
 
         else:
-            temporary_file = self.inputs.cleaned_file
+            dropped_clean_file = self.inputs.cleaned_file
```
tsalo, line 392: This fixes a recent bug that I missed in whichever PR introduced it.

PR #592, opened 2022-10-19
Output all filtered motion parameters

tsalo:
Closes None. I realized that, since we now output the filtered framewise displacement time series in a TSV with a header, there's no reason we shouldn't output the other filtered motion parameters as well.

- Include the other filtered motion parameters in the `motion.tsv` output.
- Rename `desc-framewisedisplacement_motion.tsv` to `desc-filtered_motion.tsv` or `motion.tsv`, depending on if filtering was done or not.

### DIFF
--- a/xcp_d/interfaces/qc_plot.py
+++ b/xcp_d/interfaces/qc_plot.py
@@ -283,7 +283,7 @@ class QCPlot(SimpleInterface):
 
         if self.inputs.tmask:  # If a tmask is provided, find # vols censored
             tmask_df = pd.read_table(self.inputs.tmask)
-            tmask_arr = tmask_df["framewise_displacement"]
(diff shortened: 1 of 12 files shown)


xcp_d/interfaces/qc_plot.py
```diff
@@ -283,7 +283,7 @@ def _run_interface(self, runtime):
 
         if self.inputs.tmask:  # If a tmask is provided, find # vols censored
             tmask_df = pd.read_table(self.inputs.tmask)
-            tmask_arr = tmask_df["framewise_displacement"]
+            tmask_arr = tmask_df["framewise_displacement"].values
```
tsalo, line 286: This fixes a bug introduced by recent PRs. Somehow those PRs' tests weren't catching it.

kahinimehta (approved):

PR #593, opened 2022-10-19
Fix concatenation code w.r.t. recent output changes

tsalo:
Closes None.

This will introduce a _bunch_ of merge conflicts with #522, but I'm happy to address them myself.

- Work with the new outputs in concatenation.
- Temporarily run the concatenation in all tests. I'll probably drop it down to one test before merging.
     - The nifti tests were passing (minus the output file check) as of f67d71e13f1f1920f3d40a00bea5c1a47143d9a9, so I've stopped running concatenation on those and am keeping it in the cifti test permanently.
- Refactor the concatenation code. Things like:
    - Replace `+` with os.path.join or f-strings as appropriate.
    - Abstract repeated strings into variables.
    - Simplify _some_ repeated code. `concatenate_nifti` and `concatenate_cifti` are nearly identical, so it would be good to combine them or move the repeated code into new functions at some point.


tsalo:
Nifti with FreeSurfer and Nifti without FreeSurfer both ran all the way through with the concatenation flag, so I've dropped that flag from both of those tests and am keeping it in the Cifti test.

xcp_d/utils/concantenation.py
```diff
@@ -114,22 +112,33 @@ def make_dcan_df(fds_files, name):
...
+        split_str = "_motion"
+
+    cifti_file = (
+        fds_files[0].split(split_str)[0] + '_space-fsLR_den-91k_desc-denoised_bold.dtseries.nii'
+    )
+    nifti_file = (
+        fds_files[0].split(split_str)[0] + '_space-MNI152NLin2009cAsym_desc-denoised_bold.nii.gz'
+    )
+
+    if os.path.isfile(cifti_file):
```
kahinimehta, line 129: I believe there is a _get_tr function in plot.py that does this

xcp_d/utils/concantenation.py
```diff
@@ -180,149 +191,210 @@ def concatenate_nifti(subid, fmridir, outputdir, ses=None, work_dir=None):
...
+        all_func_files = glob.glob(os.path.join(outputdir, subid, f'ses-{ses}', 'func', '*'))
+        func_dir = os.path.join(fmridir, subid, f'ses-{ses}', 'func')
+        figures_dir = os.path.join(outputdir, subid , 'figures')
 
     # extract the task list
+    denoised_bold_files = fnmatch.filter(all_func_files, '*_desc-denoised_bold.nii.gz')
     tasklist = [
-        os.path.basename(j).split('task-')[1].split('_')[0]
-        for j in fnmatch.filter(all_func_files, '*_desc-residual_bold.nii.gz')
+        denoised_bold_file.split('task-')[-1].split('_')[0]
```
kahinimehta, line 226: Is there a reason for changing this? I tried locally but either seems to work for me

xcp_d/utils/concantenation.py
```diff
@@ -180,149 +191,210 @@ def concatenate_nifti(subid, fmridir, outputdir, ses=None, work_dir=None):
...
+            glob.glob(
+                os.path.join(
+                    func_dir,
+                    f'{preproc_base_search_pattern}*_desc-brain_mask.nii.gz',
+                ),
+            ),
+        )[0]
+
+        segfile = get_segfile(preproc_files[0])
+        TR = nb.load(preproc_files[0]).header.get_zooms()[-1]
```
kahinimehta, line 328: should be a function that does this

kahinimehta (approved):

xcp_d/utils/concantenation.py
tsalo: The main reason I changed it was because `j` isn't an informative variable name. The `os.path.basename` is unnecessary because we're grabbing the split from the last instance of `task-` in the string, so it's robust to previous cases of `task-` in the full path.

kahinimehta: got it

tsalo:
Thanks for the fix @kahinimehta! Merging now.

PR #599, opened 2022-10-20
[FIX] Update load_aroma so the correct file is being read in

kahinimehta:
update datafile --> original_file in load_aroma


tsalo:
Is datafile used at all in load_confound_matrix now? If not, you can remove the parameter completely.

tsalo:
Once you drop `datafile` from `load_confound_matrix` and make the associated change in `Regress`, I think this should be good to go.

tsalo (approved):
The changes look good to me. It looks like you need to address some merge conflicts though.

tsalo:
The merge conflicts were small, so I handled them. Once CI passes, we can merge.

Issue #311, opened 2022-04-08
[RF] Concatenation


mb3152:
- [x] Make sure DCAN motion file is produced as we want
- [x] Rename 'concantenation.py' to 'concatenation.py'
- [ ] Change concatenation point to NA 
- [x] https://github.com/PennLINC/xcp_d/issues/329
- [x] Refactor the code
- [x] Test the code
- [x] Merge MVP code into main

mattcieslak:
add option for --concatenate-by

kahinimehta:
fixing concatenate_nifti

kahinimehta:
concatnate_nifti works. Next steps: change concatenation point to NA and concatenate all BOLD files tog. - on hold for now

kahinimehta:
also concatenate by task for hdf5 if not done

tsalo:
What does "Change concatenation point to NA" mean?

mattcieslak:
It's for the plots where a time derivative is displayed. The timepoint where two series are concatenated can have a big spike, but we don't want to see it plotted (I think)

tsalo:
Ah, okay. So NA is a NaN?

mattcieslak:
yes, in R speak

kahinimehta:
@tsalo what would the current status of this issue be?

tsalo:
I think we just need the NaNs between runs and some unit testing.

tsalo:
I've moved the NaN thing to #726.

PR #600, opened 2022-10-20
Use BIDSLayout in concatenation code

tsalo:
Related to #311. Stems from #522.

- Use BIDSLayout instead of globbing to select and rename files in concatenation workflow.
    - This should solve the problems with task names and concatenating across tasks.
    - It also should solve any problems with concatenating across spaces, should we choose to support multiple output spaces.
- Output the `-DCAN.hdf5` motion data file for both the concatenated data and each individual run when `--combineruns / -m` is used, instead of for concatenated data only as currently done.
- Move any non-concatenation-specific functions out of the concatenation module and into the appropriate module(s).
- Rename concantenation.py to concatenation.py.


kahinimehta (approved):

tsalo:
Thanks for the review @kahinimehta! I just want to also get @madisoth's thoughts on the PR before I merge it.

xcp_d/utils/concatenation.py
```diff
@@ -0,0 +1,452 @@
...
+                    suffix="motion",
+                    extension=".tsv",
+                    **task_entities,
+                )
+                if len(motion_files) == 0:
+                    continue
+                elif len(motion_files) == 1:
+                    # Make DCAN HDF5 file from single motion file
+                    dcan_df_file = (
+                        f"{'.'.join(motion_files[0].split('.')[:-1])}-DCAN.hdf5"
```
madisoth, line 99:
```suggestion
                        f"{'.'.join(motion_files[0].path.split('.')[:-1])}-DCAN.hdf5"
```

I got the error below on tasks with a single run; with the above change it completed w/o error and output looks good.  

My test data was the colornest CIFTI w/ FreeSurfer, with the second of the two "task-rest" runs removed (including the corresponding files in the fMRIPrep figures dir).
```
Traceback (most recent call last):
  File "/usr/local/miniconda/bin/xcp_d", line 8, in <module>
    sys.exit(main())
  File "/usr/local/miniconda/lib/python3.8/site-packages/xcp_d/cli/run.py", line 648, in main
    failed_reports = generate_reports(
  File "/usr/local/miniconda/lib/python3.8/site-packages/xcp_d/interfaces/report_core.py", line 153, in generate_reports
    concatenate_derivatives(
  File "/usr/local/miniconda/lib/python3.8/site-packages/xcp_d/utils/concatenation.py", line 99, in concatenate_derivatives
    f"{'.'.join(motion_files[0].split('.')[:-1])}-DCAN.hdf5"
  File "/usr/local/miniconda/lib/python3.8/site-packages/bids/layout/models.py", line 224, in __getattr__
    raise AttributeError("%s object has no attribute named %r" %
AttributeError: BIDSDataFile object has no attribute named 'split'
```

tsalo: Great catch! Thanks!

madisoth (approved):

Issue #566, opened 2022-10-10
[RF] Parse one output space type via flag


kahinimehta:
Order of priority: 
1. Check value of `output-space` parameter first.
    1. If there is a value, look for preprocessed derivatives in that space.
    2. If they don't exist, raise an error. 
    3. If they do, post-process the data in that space.
2. If `output-space` isn't defined, check for `MNI152NLin6Asym` space data.
    1. If there is data in this space, post-process it.
    2. If there isn't, try to warp available data to this space as part of the workflow.
    3. If data can't be warped to `MNI152NLin6Asym` space, raise an error. It would be useful for the error to include a list of available spaces, based on the preprocessed data.

We'll need to parse fMRIPrep space strings (e.g., `MNIPediatricAsym:res-1:cohort-2`) to extract all of the relevant entities needed for selecting files.

tsalo:
Regarding space string parsing, we may need [`niworkflows.utils.spaces.OutputReferencesAction`](https://github.com/nipreps/niworkflows/blob/32a33d61ab64e262a4ebcf8afcf232d70c2f92b3/niworkflows/utils/spaces.py#L686) to standardize the strings from the command-line argument, along with [`niworkflows.utils.spaces.Reference`](https://github.com/nipreps/niworkflows/blob/32a33d61ab64e262a4ebcf8afcf232d70c2f92b3/niworkflows/utils/spaces.py#L59) and [`niworkflows.utils.spaces.SpatialReferences`](https://github.com/nipreps/niworkflows/blob/32a33d61ab64e262a4ebcf8afcf232d70c2f92b3/niworkflows/utils/spaces.py#L339) to manage the references. I don't really know how the latter two classes are used yet, so I'll need to dig through the code at some point.

tsalo:
Users can now provide this information via the bids filter file, so I don't know if we need a separate flag.

tsalo:
Per yesterday's informatics call, we're fine closing this.

Issue #557, opened 2022-10-04
Preprocessing requirements are not documented


tsalo:
## Summary

In #556 a user reported a bug in which the selection of different output spaces in fMRIPrep led to different errors. We don't document anywhere which output spaces are required for `xcp_d` to run. We should also perform more validation earlier in the workflow, so that error messages are interpretable and errors don't take too long to pop up.

## Additional details

- xcp_d version: 0.1.3

tsalo:
It looks like any output spaces other than the primary MNI ones will lead to errors. For example, in #556, the problem was the existence of T1w-space BOLD outputs, even though the supported MNI ones were also present.

tsalo:
Having multiple output spaces, even if they're all supported, seems to be a problem as well.

tsalo:
The bug regarding output spaces is resolved in #603, but we should improve the documentation related to spaces and preprocessing.

Issue #350, opened 2022-06-15
Check template space to warn if parameters are appropriate


mattcieslak:
If using an adult MNI template, check that the filtering parameters are in the right range

tsalo:
Sorry, could you expand on this? What filtering parameters?

tsalo:
We can follow up in this issue if necessary, but I think things are alright as is.

PR #603, opened 2022-10-25
Select BOLD files in a single space

tsalo:
Related to #566, #557, and #350.

- Within `xcp_d.utils.bids.collect_data`, loop through allowed template spaces to find BOLD files in the most preferable available space, and only in that space. Also reduce to only include one resolution and density.
- Remove `select_cifti_bold`. `collect_data` now grabs only the files it will use.

None


tsalo:
This is a really kludgy way to select BOLD files in one space (ignoring different resolutions and/or densities). It loops through preferred spaces and grabs the files from the first one it finds.

kahinimehta (approved):

tsalo:
I think I've got the resolutions and densities handled now 🎉

tsalo:
We should really try testing on datasets with multiple resolutions and densities in the same space.

tsalo:
This PR will limit support to five spaces: MNI152NLin6Asym, MNI152NLin2009cAsym, MNIInfant, fsLR, and fsaverage. I don't know if fsaverage should be supported (only when fsLR data are not available, of course). @madisoth WDYT?

tsalo:
Dropped fsaverage, but we've added supporting it to the roadmap.

PR #606, opened 2022-10-26
Remove unused dependencies

tsalo:
Closes None.

- Sort requirements in setup.cfg, so it's easier to read.
- Remove unused requirements:
    - natsort was used in concantenation.py, but isn't used since #600 was merged.
    - networkx is unused. If we are interested in calculating graph theory metrics, we can add it back in later.
    - pytest doesn't need to be included in the main requirements. I kept it in the test requirements though.
- Remove `test_requires` section. We still have `extras_require[tests]`, and we don't use setuptools for testing, so I think it's fine. 
- Remove "duecredit" requirement group. Duecredit isn't used anywhere in xcpd.
- Remove "sentry" requirement group. Sentry-related requirements are in the main requirements list.
- Remove "resmon" requirement group. I don't know what it was for, but it was empty.
- Remove aliased "docs" requirement group. We can just use "doc".


tsalo:
Actually, it seems like it's a problem with the scipy version being installed. networkx needs 1.8.0+ to use the scipy.sparse module, but 1.6.2 is installed because of the old nilearn version.

tsalo:
Nipype requires networkx >= 2.0. Installation collects networkx 2.8.7, and the networkx repo clearly shows scipy >= 1.8.0 as a dependency for 2.8.7, so I don't know why that requirement isn't propagated to the xcpd installation... however, I think the easiest thing is just to add scipy to our requirements.

kahinimehta (approved):

tsalo:
I managed to add brainsprite and tempita back in after removing them in #607 🤦. Fixing in #635.

PR #607, opened 2022-10-26
Use Nilearn for brainsprite generation

### LINKED ISSUE #605: Try using nilearn's brainsprite utility instead of separate library
## Summary

This stems from #573. The newer versions of nilearn won't work with the brainsprite library, which hasn't been updated in a couple of years. We _could_ open an issue in the brainsprite repo (and may still have to) asking for a dependency update. However, it looks like nilearn has a function, [`view_img`](https://nilearn.github.io/stable/modules/generated/nilearn.plotting.view_img.html) that generates a brainsprite HTML snippet from a nilearn figure. We should see if that will do what we want first.

tsalo:
Closes #605.

- Remove `brainsprite` and `tempita` from requirements.
- Remove `brainsprite_template.html` from data folder.
- Increase pinned Nilearn version to 0.9.2.
- Use Nilearn's `plotting.view_img` to generate brainsprite HTML files instead of the brainsprite package.


tsalo:
nifti_without_freesurfer failed due to the following, when downloading and extracting the data from Box:

```
xz: (stdin): Compressed data is corrupt
tar: Unexpected EOF in archive
tar: Unexpected EOF in archive
tar: Error is not recoverable: exiting now
```

I'm hoping it's just a one-off, but I'm posting it here in case it's not.

kahinimehta (approved):

tsalo:
Reverted by #702.

PR #610, opened 2022-10-26
Use Nilearn for denoising

### LINKED ISSUE #573: Leverage nilearn for denoising
## Summary
Nilearn's `image.clean_img` and `signal.clean` can perform the following steps in one pass:

1. Temporal masking/censoring
2. Bandpass filtering with butterworth or cosine filters.
3. Confound regression.
4. Detrending.

Also, NiftiLabelsMasker can apply the same steps, along with smoothing.

`interfaces.fmriprep.load_confounds` can also load fMRIPrep confounds according to different denoising strategies (i.e., parameter sets).

The one problem is that nilearn doesn't play well with CIFTI data yet, so we still need Connectome WB. I'd like to push as much Python CIFTI I/O work up to nilearn as we can though.

tsalo:
Closes #573. We will need to wait until nilearn 0.9.3 before we can merge this.

Nilearn doesn't let you set the filter order (`bpf_order`). Should we just drop that argument?
Nilearn uses the default filtfilt parameters automatically, so we cannot use an even padtype or a custom padlen.

- Replace the censoring, interpolation, filtering, and denoising steps with nilearn.

None, yet.


tsalo:
The NiftiWithoutFreeSurfer dataset is too short (16 vols) for bandpass filtering with the default (i.e., nilearn's hardcoded) padlen (33 vols). I tried to disable the bandpass filter in this test, but that led to #615.

Also, combining tmask and dummyvols doesn't play well with QCPlot.

tsalo:
How important is the post-regression, pre-filtering DVARS line in the executive summary carpet plot?

tsalo:
I will need to hold off on this until https://github.com/PennLINC/xcp_d/pull/779 is handled and a new nilearn release is made.

PR #602, opened 2022-10-24
[ENH] Support Nibabies ingression into XCP

kahinimehta:
- Pull data from the newly created PennBox folders
- Fix bug with non-binary masks from Nibabies data
- Add tests for Nibabies

PR #619, opened 2022-10-28
Revert removed code from nibabies PR

tsalo:
Adds back in the resolution/density filters to `collect_data`, removed in #602.


tsalo:
Going to merge this but we need to start logging more in concatenation.py to make sure we know why CI is breaking.

PR #621, opened 2022-10-28
Use BIDSLayout instead of globbing functions to collect necessary files

tsalo:

- Move `xcp_d.utils.plot._get_tr` to `xcp_d.utils.bids._get_tr`. This is partially to prevent circular imports, but also it just makes more sense in that module.
- Remove unused `xcp_d.utils.filemanip.simplify_list` function.
- Add `t1w_mask `, `t1w_seg `, `t1w_to_mni_xform`, and `mni_to_t1w_xform` to `xcp_d.utils.bids.collect_data`.
- Create a new function, `xcp_d.utils.bids.collect_run_data`, to collect the files necessary for a single BOLD file. This grabs the associated confounds file, metadata, reference image, brain mask, and T1w-to-native form.


kahinimehta (approved):

tsalo:
I want to show the collected files in a logger message, but for some reason the message refuses to show up. I'm stumped 😕

tsalo:
It looks like the nipype logger is set to a given, user-requested level at:

https://github.com/PennLINC/xcp_d/blob/97047efdabddb0d0371eb977c844d96e6856da08/xcp_d/cli/run.py#L520-L522

But that level is ignored within the xcpd workflows, because level is reset back to 20 (info) automatically around:

https://github.com/PennLINC/xcp_d/blob/97047efdabddb0d0371eb977c844d96e6856da08/xcp_d/cli/run.py#L686

tsalo:
I moved the `setLevel` calls to within `build_workflow`, and they seem to work now. The logged messages will include full paths and metadata, so I didn't want them to be shown at the INFO level. I think the DEBUG level is reasonable.

xcp_d/cli/run.py
```diff
@@ -889,7 +889,13 @@ def build_workflow(opts, retval):
...
         {
-            "logging": {"log_directory": str(log_dir), "log_to_file": True},
+            "logging": {
+                "log_directory": str(log_dir),
+                "log_to_file": True,
+                "workflow_level": log_level,
+                "interface_level": log_level,
+                "utils_level": log_level,
+            },
             "execution": {
```
tsalo, line 899: I'm not sure this has any effect, but it seems like the right thing to do?

PR #629, opened 2022-11-01
Cache the downloaded test data

### LINKED ISSUE #337: [BUG] Example data on circleci is too big
It looks like there are 3 copies of a full freesurfer dataset in the example data, which makes it take a long time to download.

tsalo:
Closes #337.

- Eliminate duplicated test data by downloading the data in a step _before_ the actual tests.
- Keep data saved in a long-term cache.

None

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,6 +33,58 @@ jobs:
       - checkout
       - run: *runinstall
 
+  download_data_fmriprepwithfreesurfer:
+    <<: *dockersetup
+    working_directory: /src/xcp_d/.circleci/
+    steps:
+      - checkout
+      - restore_cache:
+          key: fmriprepwithfreesurfer-01
+      - run:
+          TESTDIR="/src/xcp_d/.circleci"
+          cd ${TESTDIR}
+          source get_data.sh
+          get_bids_data ${TESTDIR} fmriprep_colornest
+          get_bids_data ${TESTDIR} freesurfer_colornest
+      - save_cache:
+          key: fmriprepwithfreesurfer-01
+          paths:
+              - /src/xcp_d/.circleci/data/fmriprepwithfreesurfer
+
+  download_data_fmriprepwithoutfreesurfer:
+    <<: *dockersetup
+    working_directory: /src/xcp_d/.circleci/
+    steps:
+      - checkout
+      - restore_cache:
+          key: fmriprepwithoutfreesurfer-01
+      - run:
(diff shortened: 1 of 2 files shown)


kahinimehta (approved):

mattcieslak:
This looks spot on. Have you noticed an improvement in performance?

mattcieslak:
We might actually want to use workspaces so we don't need to keep the data on circleci between workflow runs: https://circleci.com/blog/persisting-data-in-workflows-when-to-use-caching-artifacts-and-workspaces/

tsalo:
>This looks spot on. Have you noticed an improvement in performance?

I'm doing a head-to-head comparison with `main` now, so I'll update this comment when I have the actual gain.

>We might actually want to use workspaces so we don't need to keep the data on circleci between workflow runs: https://circleci.com/blog/persisting-data-in-workflows-when-to-use-caching-artifacts-and-workspaces/

That would definitely reduce the amount of data, but it wouldn't speed up the tests, since data would need to be downloaded at the beginning of each workflow run. Is the data storage a major concern? I thought there was room in our plan to cache some data.

mattcieslak (approved):

mattcieslak:
I think the only gotcha here is that if we want to change the input data we should change the cache key

tsalo:
Would it make sense to make the Docker image, with xcpd installed, workspace data?

mattcieslak:
The jobs are running inside a docker image of xcpd base and xcpd gets installed as part of the job itself. Caching and loading the docker image takes ~40 minutes on each end and takes up a ton of cache storage. Circleci emailed me back when they changed their policy to warn me to stop doing that.

From what I understand, circleci has its own network that sends data to and from instances that run the docker jobs. Dockerhub is a part of this network and we don't get charged if they have to download xcpd_build from dockerhub and we then install xcpd on top of it. It also downloads xcpd_build very quickly, much more quickly than restoring the docker image from an explicitly-named cache and unzipping it

tsalo:
Ah, okay. At the moment, the `build` step doesn't actually do anything for the later steps, since it's not passing along the downloaded Docker image with xcpd installed, and it takes ~3.5 minutes to run. What if I just dropped that step?

mattcieslak:
That step is the gateway step, where if the pip install doesn't work it prevents the other jobs from running (as there's no way they could succeed). They should probably also be prevented from running if the data download fails

tsalo:
Good point. I'll keep it in then.

PR #630, opened 2022-11-01
Use nearest neighbor interpolation in ribbon brainsprites

tsalo:


tsalo:
Closing this PR since we probably won't use this solution. A decision will probably be made at Friday's meeting.

PR #623, opened 2022-10-31
Simplify transform-getting functions

tsalo:
Closes None.

The transform lists seem to be a mix for forward and backward transforms, and I'm not sure they're always in the right order.

## Changes proposed in this pull request

- Rename `get_transformfilex` to `get_bold2std_and_t1w_xforms`.
- Rename `get_transformfile` to `get_std2bold_xforms`.
- Attempt to document and simplify `get_std2bold_xforms` and `get_bold2std_and_t1w_xforms`.
- Remove spaces that xcpd doesn't support.
- Rename the locally-stored MNIInfant-to-MNI152 and back transforms.

PR #632, opened 2022-11-02
Fix MNI152NLin6Sym import in utils.py

smeisler:
I presumed the original duplicate MNI152NLin6Asym template finding was a typo.

Fix utils.py to accept MNI152NLin6Sym

N/A


tsalo:
Thanks for opening this, but #623 will substantially refactor that function and address the problem.

smeisler:
Got it, sounds good!

Issue #419, opened 2022-08-18
Add and test custom confounds utilities to overall workflow


kahinimehta:
Adding to workflows:

- [x] Dummytime
- [x] Censorscrub
- [x] Regression
- [x] QC
- [x] Executive summary

Testing

- [x] Dummytime
- [x] Censorscrub
- [x] Regression
- [x] QC
- [x] Executive summary

tsalo:
Look at PNC n-back data for testing.

PR #642, opened 2022-11-03
Support custom confounds with headers

tsalo:
Related to #419. One thing that is untested is the code to actually collect the confounds files.

- ~~Allow users to provide a single custom confounds file.~~
- Simplify handling of custom confounds.
- Require headers in custom confounds files. This is the breaking change.
- Expect custom confounds files to have the same name as fMRIPrep confounds files. This is another breaking change.
- Add a test for using custom confounds.
- Create a design matrix figure.

I updated the custom confounds section of the usage page.


.circleci/tests/test_confounds.py
```diff
@@ -0,0 +1,72 @@
...
+import os
+
+import numpy as np
+import pandas as pd
+from nilearn.glm.first_level import make_first_level_design_matrix
+
+from xcp_d.utils.confounds import load_confound_matrix
+
+
+def test_custom_confounds(data_dir, tmp_path_factory):
```
tsalo, line 11: We can use this basic approach for testing out the different confound strategies.

kahinimehta (approved):

tsalo:
@mattcieslak WDYT about letting users provide a single custom confounds file? I could add a check to raise an Exception if the call grabs more than one BOLD file to post-process, so it only takes effect if there's one run and users know that their custom confounds are associated with that run.

tsalo:
Per today's SCRUM, I'll drop the single-file support.

Issue #551, opened 2022-10-03
Harmonize atlases across XCP, QSIPrep


kahinimehta:
Duplicate of https://github.com/PennLINC/informatics_misc/issues/49
- [x] Arrange meeting with @mattcieslak and @tsalo 
- [x] Check xcpd surface atlases against qsiprep volumetric atlases.
- [ ] Make atlases more templateflow-y/BIDS-y in Box folders.
- [ ] Ensure discrete segmentation atlases have data in int16.
- [ ] Add surface atlases to Box folders.
- [ ] Build atlases into Docker image, using Box URLs.
- [ ] Remove atlases xcp_d data folder.
- [x] Create time series files for the different subcortical atlases separately.
- [x] Create specific combined connectivity matrices between subcort and cort atlases based on qsiprep's outputs.
- [ ] Try to back-prune the data folder in git history to reduce size of the repository. (low priority)
- [x] @kahinimehta: check the order of regions, then map the surface-based XCP regions to the QSIprep order if needed (crossref the atlas config file in QSIprep and .dlabel/ node order info in XCP) - don't seem to be matching up at the moment....

tsalo:
Notes from meeting today:

1. Upload NLin6-1x1x1 versions of all atlases we want to the Box xcpd folder.
2. Upload fsLR-32k versions of all atlases, preferably with separate files for cortical and subcortical atlases to the same folder. This may require some work to get the subcortical atlases in this format.
3. Come up with a good format for metadata, node labels, and network labels. qsiprep currently uses a single JSON file for all atlases, xcpd uses text files (.1D),  templateflow uses TSV files. Basically, there's room to standardize. The BIDS Atlas BEP is very unstable and may not be worth adhering to.
4. Write code to do the following:
    1. Combine NLin6 cortical and subcortical atlases, per qsiprep's requirements.
    2. Combine the associated metadata and label files, as needed.
    3. Warp the combined NLin6 nifti atlases to 2009c space.
    4. ~~Enforce LPS+ orientation.~~
    5. Write the new expanded atlases and their sidecar files to the qsiprep Box folder.
10. Upload the code to the code Box folder.
11. Update the Box URL in the qsiprep Docker image.
12. Update the xcpd Docker image to download the atlases from the xcpd Box folder.
13. Remove the atlases from the xcpd repo.
14. Update xcpd to combine atlases as needed. This may require some work to use the new subcortical CIFTI atlases.

kahinimehta:
[...]

PR #647, opened 2022-11-04
Output warped atlases to derivatives

### LINKED ISSUE #643: Copy atlases to top of derivatives folder
## Summary
I don't think the actual atlases are accessible to users at the moment, so we should add copies to the top of the xcpd output directory.

## Additional details
Users get the images and TSV files if they want to unmask/visualize their data, or even just to get relevant metadata/labels.

## Next steps
1. Add a node in the atlas-fetching workflow that checks if the atlas files exist in the output directory and, if not, copies them over.

tsalo:
Closes #643.

- Replace the copied version of DerivativesDataSink with a child class inheriting from the official niworkflows version.
- Pin niworkflows to 1.7.1.
- Add a step to downcast the CIFTI atlases to int16 (cast_cifti_to_int16), because niworkflows's DerivativesDataSink will attempt to treat dlabel.nii files as NIFTIs. Trying to set the data type of a CIFTI using the same command as one would for a NIFTI will cause an error. We circumvent that by ensuring that the files have the appropriate data type ahead of time.
- Copy atlas files to top level of xcpd derivatives folder, so users have access to the atlases. We will need to update this when we standardize our atlas format (e.g., by copying over the TSVs and JSONs as well).

None


kahinimehta (approved):

kahinimehta:
Might have to update the fcon tests @tsalo

tsalo:
Note about c6c5a1e719ed3cbdcede4ddfe6dc68719bc8ca6b- our copy of DerivativesDataSink tries to coerce dseg files to int16, which fails on CIFTI files. However, the niworkflows version _should_ work on CIFTIs, so it seems like we should use that instead.

tsalo:
It looks like DerivativesDataSink doesn't provide any way to override the default path patterns and entities, so I think I'll need to copy over the source code. Given that niworkflows was relicensed since the last time we did this, I will need to look at their licensing and see what needs to be done to incorporate modified versions of their code into xcpd.

tsalo:
This may need to wait until https://github.com/nipreps/niworkflows/issues/757 is resolved.

tsalo:
I should be able to use niworkflows' DerivativesDataSInk once https://github.com/nipreps/niworkflows/pull/762 is merged and a new release is made. Then we just import it, use it as a parent class for a new DerivativesDataSink, and override some of the attributes.

tsalo:
I'm going to merge this now. I think we will need to also output the sidecar files and node labels once #551 is addressed.

PR #648, opened 2022-11-04
Output T2w images if available

### LINKED ISSUE #644: Output preprocessed T2w images along with the T1w images
Just make sure it's in the QC- copy it into the output directory, in the same space as T1

tsalo:
Closes #644. I'm going to need test data with T2w images before I can merge this.

To do:
- [ ] Add nibabies test data with processed T2w image.
- [x] Ensure that the T2w is outputted.

- Select preprocessed T2w image in `collect_data`. This image is optional, so do not raise an error if no image is found.
- Output T2w image (after warping to standard space, if necessary) in `init_t1w_wf`.


kahinimehta (approved):

tsalo:
I think we should use nibabies derivatives to test this, so it may depend on https://github.com/nipreps/nibabies/issues/88.

PR #654, opened 2022-11-07
Map abbreviated entities to full names for templateflow get call

### LINKED ISSUE #653: TemplateFlow has different entities in their BIDSLayout config
This is causing ReHo/ALFF plotting to throw an error - we need to map between different versions of the entity names; "resolution" instead of "res", "density" instead of "den", and "description" instead of "desc".

tsalo:
Closes #653.

Templateflow's BIDSLayout config file uses "density" instead of "den" and "resolution" instead of "res". Other entities are the same between templateflow and pybids/xcpd.

- Map "den" to "density" and "res" to "resolution" for templateflow get call.

None


tsalo:
Going to merge this.

PR #660, opened 2022-11-10
Remove dummy volumes from custom confounds files

tsalo:
In RemoveTR, the step that removes dummy volumes from custom confounds files was actually only _retaining_ the dummy volumes, instead of dropping them.

Drop dummy volumes, instead of keeping _only_ dummy volumes, in custom confounds file.


tsalo:
The bug was introduced by me in #642.

tsalo:
Unfortunately, we don't have any workflow tests with custom confounds, so I will just merge as-is.

PR #661, opened 2022-11-10
Remove dummy volumes from beginning of each run in concatenation workflow

tsalo:
Currently, the concatenation workflow removes dummy volumes from the beginning of the concatenated preprocessed BOLD data within `plot_svgx`. This misses the dummy volumes at the beginning of each run after the first, as they are buried in the middle of the concatenated time series.

- Add a `dummy_scans` parameter to `_concatenate_niimgs`.
- Create a new function, `_drop_dummy_scans`, to remove dummy volumes from BOLD files. This will be used by both `RemoveTR` (in the main workflow) and `_concatenate_niimgs` (in the concatenation workflow, with `--dcan-qc` enabled).
- Only concatenate preprocessed files, calculate preprocessed and post-processed data DVARS, and flag segmentation and mask files, if `dcan_qc` is enabled.
- Remove dummy volumes from preprocessed DVARS.
- Mask the pre- and post-processed DVARS with the outlier mask.

### DIFF
--- a/.circleci/CiftiWithFreeSurferTest.sh
+++ b/.circleci/CiftiWithFreeSurferTest.sh
@@ -40,4 +40,5 @@ $XCPD_CMD \
     --cifti \
     --combineruns \
     --dcan-qc \
-    --dummytime 8
+    --dummytime 8 \
+    --fd-thresh 0.04
--- a/.circleci/CiftiWithFreeSurferTest.sh
+++ b/.circleci/CiftiWithFreeSurferTest.sh
@@ -31,10 +31,13 @@ XCPD_CMD=$(run_xcpd_cmd ${BIDS_INPUT_DIR} ${OUTPUT_DIR} ${TEMPDIR})
 echo $XCPD_CMD
 
 $XCPD_CMD \
-    --despike --head_radius 40 \
-    --smoothing 6 -vvv \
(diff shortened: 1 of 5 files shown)


tsalo:
Here is the error before the fix:

```
ValueError: Shapes do not match:
	/tmp/tmpbkfjn3x5/rawdata.dtseries.nii: (91282, 364)
	/src/xcp_d/.circleci/cifti_with_freesurfer/derivatives/xcp_d/sub-colornest001/ses-1/func/sub-colornest001_ses-1_task-rest_space-fsLR_den-91k_desc-denoised_bold.dtseries.nii: (91282, 360)
	/src/xcp_d/.circleci/cifti_with_freesurfer/derivatives/xcp_d/sub-colornest001/ses-1/func/sub-colornest001_ses-1_task-rest_space-fsLR_den-91k_desc-denoised_bold.dtseries.nii: (91282, 360)
```

Essentially, the dummy volumes (4 volumes for a dummytime of 8 seconds and a TR of 2 seconds) are removed from the beginning of the concatenated preprocessed BOLD file, resulting in 364 (184 * 2 - 4) volumes, while the concatenated denoised BOLD file, which had 4 dummy volumes removed before concatenation, has 360 volumes.

PR #675, opened 2022-11-18
Load confounds via Nilearn functionality

### LINKED ISSUE #671: Leverage nilearn for confound selection
## Summary
Let's take a hard look at [`nilearn.interfaces.fmriprep.load_confounds`](https://nilearn.github.io/stable/modules/generated/nilearn.interfaces.fmriprep.load_confounds.html) to see if we can use it for confound selection.

## Additional details
It would simplify the xcp_d codebase, and could open up more confound strategies to users.

## Next steps

1. Compare load_confounds confound selection options to options in xcp_d.
2. Use load_confounds if it supports everything xcp_d does.

kahinimehta:
Closes #671 

Replace XCP's confounds functionality with Nilearn.

### DIFF
--- a/xcp_d/utils/confounds.py
+++ b/xcp_d/utils/confounds.py
@@ -200,7 +200,7 @@ def load_confound_matrix(params, img_file, custom_confounds=None):
     elif params == "acompcor_gsr":  # Get the rot and trans values, as well as their derivative,
         # acompcor and cosine values as well as global signal
         confound = load_confounds(img_file, strategy=(['motion', 'high_pass', 'compcor',
-                                                        'global_signal']),
+                                                       'global_signal']),
                                   motion='derivatives', compcor='anat_separated',
                                   global_signal='basic', n_compcor=5)[0]
     elif params == "custom":
@@ -213,25 +213,6 @@ def load_confound_matrix(params, img_file, custom_confounds=None):
         confound = pd.concat([custom, confound], axis=1)
 
     return confound
-
-
-def load_aroma(confounds_df):
-    """Extract aroma confounds from a confounds TSV file.
-
-    Parameters
-    ----------
-    confounds_df : :obj:`pandas.DataFrame`
-        The confounds DataFrame.
-
-    Returns
-    -------
-    pandas.DataFrame
-        The AROMA noise components.
-    """
-    aroma_noise = [c for c in confounds_df.columns if c.startswith("aroma_motion_")]
-    aroma = confounds_df[aroma_noise]
-
-    return aroma
 
 
 @fill_doc
--- a/xcp_d/utils/confounds.py
+++ b/xcp_d/utils/confounds.py
@@ -181,40 +181,33 @@ def load_confound_matrix(params, img_files, custom_confounds=None):
     can cause regression to happen incorrectly.
     """
     if params == "24P":  # Get rot and trans values, as well as derivatives and square
-        confound = load_confounds(img_files, strategy=(['motion']), motion='full',
-                                  scrub=100, fd_threshold=100, std_dvars_threshold=100)[0]
+        confound = load_confounds(img_files, strategy=(['motion']), motion='full')[0]
     elif params == "27P":  # Get rot and trans values, as well as derivatives and square, WM, CSF,
         # global signal
         confound = load_confounds(img_files, strategy=(['motion', 'wm_csf', 'global_signal']),
-                                  scrub=100, fd_threshold=100, std_dvars_threshold=100,
                                   motion='full')[0]
     elif params == "36P":  # Get rot and trans values, as well as derivatives, WM, CSF,
         # global signal, and square. Add the square and derivative of the WM, CSF
         # and global signal as well.
         confound = load_confounds(img_files, strategy=(['motion', 'wm_csf', 'global_signal']),
-                                  scrub=100, fd_threshold=100, std_dvars_threshold=100,
                                   motion='full', global_signal='full', wm_csf='full')[0]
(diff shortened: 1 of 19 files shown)


kahinimehta:
@tsalo Here were some of the questions I had: 

1. I can’t get acompcor to work - even though I supply “high_pass”, which is supposed to create cosine values,  it can’t find the cosine values - could this be because it requires 1.2.X or above from fMRIPrep? 
2. Similarly, I’m not able to test AROMA so am not 100% sure it works. 
3. "compcor" confounds derived from CompCor 1. When using this noise component, "high_pass" must also be applied. Associated parameter: compcor, n_compcor. Will this be an issue since we are motion-filtering? I’m not entirely sure what this means.
4. ‘scrub=100, fd_threshold=100, std_dvars_threshold=100’  should prevent any volumes from being scrubbed, right?
5. Should n_compcor be 5?

tsalo:
> 1. I can’t get acompcor to work - even though I supply “high_pass”, which is supposed to create cosine values,  it can’t find the cosine values - could this be because it requires 1.2.X or above from fMRIPrep?

Perhaps. Which dataset(s) does aCompCor fail on?

> 2. Similarly, I’m not able to test AROMA so am not 100% sure it works.

Makes sense. We'll need some AROMA data to make sure it's good. In the meantime, following the load_confounds documentation is probably enough.

> 3. For using compcor, I have to enable the "high-pass" parameter. `1. When using this noise component, "high_pass" must also be applied. Associated parameter: compcor, n_compcor.` Will this be an issue since we are motion-filtering? I’m not entirely sure what this means.

I thought our code already grabbed the cosine columns too? Looking at `load_cosine`, it seems like our code would actually just fail silently (i.e., return no cosine columns) when cosine columns aren't available. Raising an error seems preferable to me.

> 4. `scrub=100, fd_threshold=100, std_dvars_threshold=100` should prevent any volumes from being scrubbed, right?

Can you use None instead of 100? Although... AFAICT, none of those should be used if the strategy doesn't include "scrub", right?

> 5. Should n_compcor be 5?

I think that's right.

kahinimehta:
1. I used fmriprepwithoutfreesurfer
4. Will do - and yes, I think that's right

xcp_d/utils/confounds.py
```diff
@@ -156,168 +157,16 @@ def load_motion(
...
-    new_columns = [c + "_sq" for c in columns]
-    squared_confounds = confound**2
-    squared_confounds.columns = new_columns
-    return squared_confounds  # Square the confound data
-
-
 @fill_doc
-def load_confound_matrix(params, confound_tsv, custom_confounds=None):
-    """Load a subset of the confounds associated with a given file.
+def load_confound_matrix(params, img_files, custom_confounds=None):
```
tsalo, line 161:
```suggestion
def load_confound_matrix(params, img_file, custom_confounds=None):
```

Rather than call the parameter "img_files", which implies that it's a list of multiple files, I think you should say "img_file".

tsalo:
> 1. I used fmriprepwithoutfreesurfer

The dataset description says that it's from fMRIPrep version 20.2.1, but my guess is that's wrong. I don't think there's anything wrong with requiring a minimum fMRIPrep version, though that's something we should discuss with the team. This is related to #557.

xcp_d/utils/confounds.py
```diff
@@ -156,168 +157,16 @@ def load_motion(
...
-    """Load a subset of the confounds associated with a given file.
+def load_confound_matrix(params, img_file, custom_confounds=None):
+    """
+    Load a subset of the confounds associated with a given file.
 
     Parameters
     ----------
     %(params)s
-    confound_tsv : str
-        The path to the confounds TSV file.
```
tsalo, line 320: Given that this no longer uses the confounds file, I think we should (1) use this function to select the confounds file, rather than anything earlier in the workflow and (2) use this function to load the necessary motion parameters for CensorScrub, probably in a separate file.

xcp_d/utils/confounds.py
```diff
@@ -331,85 +180,34 @@ def load_confound_matrix(params, confound_tsv, custom_confounds=None):
     Switching the order of the trans and rot values in the motion columns
     can cause regression to happen incorrectly.
```
tsalo, line 181: Probably should remove this comment.

xcp_d/utils/confounds.py
kahinimehta: I think we have another function that does (1) and (2)?

tsalo: Exactly, those can be removed!

kahinimehta: AHHH, I see - should that be a separate PR, or should I tackle it in this one?

tsalo: Happy to leave it for a separate PR, but it would be great if you could open an issue about it at some point before we finish up this one.

kahinimehta: Yes, will create an issue right now and push a new version of this without [skip ci] - but it should be passing tests now.

tsalo (changes requested):
There are some minor changes, but what's weird is that flake8 isn't flagging things it should.

kahinimehta:
Hmm @tsalo It's failing the linting check with `xcp_d/utils/confounds.py:179:35: BLK100 Black would make changes.` , but won't say what I should change

tsalo:
@kahinimehta you should just run black on the package locally to fix that.

tsalo (approved):
Everything looks good to me, but I think we need to bump up the priority on testing the confounds. I've opened #678 about this.

Issue #684, opened 2022-11-23
Error using `--dcan-qc` and `--cifti` on nibabies data with multiple output spaces


lundq163:
## Summary
when running xcpd 0.3.0 on nibabies 22.1.2 derivatives, xcpd seems to be trying to use the wrong T1w transform file from the nibabies anat output. I was using the --dcan-qc flag when this happened; when I run xcpd on the same nibabies derivatives without the flag, then it will process without errors

## Additional details

- xcp_d version: 0.3.0

### What were you trying to do?
test --dcan-qc and --warp-surfaces-native2std flags
### What actually happened?
error produces when using --dcan-qc flag 
## Reproducing the bug
Full .err file output:

```
++ Final vox neighbood: ellipsoid with Nneigh=27 and radii (1.900,1.900,1.900).
++ ReHo (Kendall's W) calculated.
xcp_d failed: Traceback (most recent call last):
  File "/usr/local/miniconda/lib/python3.8/site-packages/nipype/pipeline/plugins/multiproc.py", line 67, in run_node
    result["result"] = node.run(updatehash=updatehash)
  File "/usr/local/miniconda/lib/python3.8/site-packages/nipype/pipeline/engine/nodes.py", line 527, in run
    result = self._run_interface(execute=True)
  File "/usr/local/miniconda/lib/python3.8/site-packages/nipype/pipeline/engine/nodes.py", line 645, in _run_interface
    return self._run_command(execute)
  File "/usr/local/miniconda/lib/python3.8/site-packages/nipype/pipeline/engine/nodes.py", line 771, in _run_command
    raise NodeExecutionError(msg)
nipype.pipeline.engine.nodes.NodeExecutionError: Exception raised while executing Node get_std2native_transform.

Traceback:
	Traceback (most recent call last):
	  File "/usr/local/miniconda/lib/python3.8/site-packages/nipype/interfaces/base/core.py", line 398, in run
	    runtime = self._run_interface(runtime)
	  File "/usr/local/miniconda/lib/python3.8/site-packages/nipype/interfaces/utility/wrappers.py", line 142, in _run_interface
	    out = function_handle(**args)
	  File "<string>", line 52, in get_std2bold_xforms
	ValueError: Transform does not match BOLD space: MNIInfant != /data/sub-patient01/ses-01/anat/sub-patient01_ses-01_from-MNI152NLin6Asym_to-T1w_mode-image_xfm.h5

Traceback (most recent call last):
[...]

PR #688, opened 2022-11-28
Use input_type to determine order of preferred spaces

tsalo:
Partially addresses #684.

- Set MNIInfant as the highest-priority space for nibabies volumetric data.
- Remove `xcp_d.workflow.execsummary.t1_to_native` in favor of `xcp_d.utils.utils.t12native`. The code is the same.
- Remove unnecessary `mni_to_t1w` input from `init_ciftipostprocess_wf` to `init_qc_report_wf`. This input is only used for non-CIFTI data, so there's no point in feeding it in from the CIFTI workflow.

None


tsalo:
I am not sure how to deal with multiple MNIInfant cohorts, since `cohort` is an entity in the preprocessed files, but not in the transforms. In the transforms, the cohort number is incorporated into the template name as `MNIInfant+<cohort>`.

tsalo:
Going to merge this without a review, but @madisoth will be testing it on his own data to make sure it solves the bug.

PR #689, opened 2022-11-28
Infer volumetric space from transform in executive summary with cifti data

tsalo:
Partially addresses #684, along with #688.

- Use the `from` entity in the `mni_to_t1w` transform filename to determine the template space (and possibly cohort) when selecting the appropriate nifti BOLD file and BOLD reference file for the executive summary. This should only take effect when using both `--cifti` and `--dcan-qc`, and should only _matter_ when there are multiple output spaces in the derivatives.
- Break up a number of function calls in the executive summary workflow into nodes.
- Create a new function, `_find_nifti_files`, that grabs the appropriate nifti files.

None.


tsalo:
Going to merge this without a review, but @madisoth will be testing it on his own data to make sure it solves the bug.

PR #690, opened 2022-11-28
Use appropriate intent codes for cifti outputs

### LINKED ISSUE #687: Appropriate CIFTI intent codes are not assigned based on extension
## Summary

@valeriejill noticed that `.dscalar.nii` images produced by XCPD 0.3.0 have 3002 intent codes, which are for `.dtseries.nii` files.

We probably need to check intent codes in the niimg objects and overwrite them for new images as necessary.

## Additional details

- xcp_d version: 0.3.0

tsalo:
Closes #687.

- Collect cifti extensions and associated intent codes from https://www.nitrc.org/forum/attachment.php?attachid=333&group_id=454&forum_id=1955, and add these to a dictionary in `write_ndata`.
- Select the appropriate cifti intent code base on the output filename's extension in `write_ndata`.

None

### DIFF
--- a/.circleci/tests/test_ALFF.py
+++ b/.circleci/tests/test_ALFF.py
@@ -165,7 +165,7 @@ def test_cifti_alff(data_dir, tmp_path_factory):
         (
             "compute_alff_wf/alff_compt/sub-color"
             "nest001_ses-1_task-rest_run-2_space-fsLR_den-91k_"
-            "bold_alff.dtseries.nii"
+            "bold_alff.dscalar.nii"
         )
     )
     original_alff_data_mean = nb.load(original_alff).get_fdata().mean()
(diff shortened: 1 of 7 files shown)


tsalo:
`CiftiSmooth` (`wb_command -cifti-smoothing`) automatically sets the intent to ConnDenseSeries, because apparently it doesn't expect a statistical image. I don't know if it's worth it to rewrite the generated file with the appropriate intent code. @valeriejill WDYT?

tsalo:
Look at post-run hook to fix the CiftiSmooth output's header.

tsalo:
CopyHeaderInterface doesn't work with ciftis... going to need to find another solution.

tsalo:
Adding a `_post_run_hook` has somehow led to bus errors, and I have no clue why. @mattcieslak have you come across this before?

tsalo:
Using a separate node works.

PR #693, opened 2022-11-29
Load basic AROMA confounds

tsalo:
Nilearn's load_confounds requires users to provide the `desc-smoothAROMAnonaggr_bold` file in order to use the "full" ICA AROMA confounds. Since we provide the regular, preprocessed BOLD file, the "full" confounds will fail.

_However_, by loading the AROMA motion components and using those in our design matrix, we are performing "aggressive" denoising, which is not good. Basically, this PR fixes a loading bug, but commits (temporarily) to a subpar denoising strategy. I plan to switch to another strategy (e.g., non-aggressive or orthogonalization) in a future PR.

- Change the ica_aroma regressor flag from "full" to "basic".

## Documentation to review

generalworkflow.rst

### DIFF
--- a/docs/generalworkflow.rst
+++ b/docs/generalworkflow.rst
@@ -92,6 +92,17 @@ Processing Steps
 
     For more information about confound regressor selection, please refer to :footcite:t:`benchmarkp`.
 
+    .. warning::
+
+      In XCP-D versions prior to 0.3.1, the selected AROMA confounds were incorrect.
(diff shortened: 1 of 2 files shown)


tsalo:
Going to test locally on data with AROMA. If it passes, I'll merge.

tsalo:
This worked locally, so I'm going to merge.

PR #695, opened 2022-11-29
Select best volumetric space even when using `--cifti` flag

tsalo:
Related to #684.

- When using the `--cifti` flag, find the best volumetric space for the `template_to_t1w_xform` and `t1w_to_template_xform` files based on both the preferred spaces and what nifti BOLD files are available. This way, the transforms won't default to MNI152NLin6, the transforms for which are generated automatically, even when it's not an output space. In situations where it's _not_ an output space, the transforms for it were being selected, but then the volumetric BOLD file in another space was selected elsewhere, raising an exception.
- Tangentially, I also fixed a bug in `collect_data` where "bold"["space"] in the filter file was not respected (and indeed raised an exception).

None.

### DIFF
--- a/xcp_d/utils/bids.py
+++ b/xcp_d/utils/bids.py
@@ -234,10 +234,6 @@ def collect_data(
             )
             if bold_data:
                 queries["bold"]["space"] = space
-                if not cifti:
-                    queries["t1w_to_mni_xform"]["to"] = space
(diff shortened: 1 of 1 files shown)


tsalo:
Running niftis + filtering based on space works for me locally. I still need to deal with cifti.

PR #697, opened 2022-11-30
Enable non-aggressive denoising with signal regressors

### LINKED ISSUE #694: Use non-aggressive denoising with AROMA options
## Summary
Currently, xcpd performs "aggressive" denoising with AROMA noise components (i.e., noise components are regressed from the data). However, "aggressive" denoising is not a good choice, and most users probably expect "non-aggressive" denoising.

We should implement non-aggressive denoising when `aroma` or `aroma_gsr` is selected.

## Additional details
Better denoising with AROMA confounds.

## Next steps
1. In the load_confound_matrix function, instead of using load_confounds for AROMA confounds, find the AROMA motion components index file and the MELODIC mixing matrix file, and load both to split the MELODIC ICA components into "noise" and "signal" groups.
2. Place the noise and signal groups in the confounds DataFrame. The signal components will need to have some kind of prefix or standardized naming convention, so that later steps don't try to regress them out of the data. fMRIPrep calls noise components "aroma_motion_[X]" in its confounds file, so maybe "aroma_signal_[X]"?
3. Implement non-aggressive denoising in the regression step. See https://tedana.readthedocs.io/en/latest/denoising.html#remove-noise-correlated-fluctuations-that-aren-t-correlated-with-fluctuations-in-accepted-components-non-aggressive-denoising for example code.

tsalo:
Closes #694.

- Mean-center and detrend confounds to match the processing done to the BOLD data, prior to regression.
- Adopt non-aggressive denoising strategy by (1) loading signal AROMA components into design matrix along with noise ones, (2) fitting GLM to all regressors (including signal components), and (3) removing the fitted _nuisance_ regressors from theta to denies it.
- Convert if/else statements in `load_confound_matrix` and `describe_regression` to dictionaries of keyword arguments.

- https://xcp-d--697.org.readthedocs.build/en/697/usage.html#signal-confounds-for-non-aggressive-denoising

### DIFF
--- a/xcp_d/interfaces/regression.py
+++ b/xcp_d/interfaces/regression.py
@@ -1,6 +1,7 @@
 # emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: nil -*-
 # vi: set ft=python sts=4 ts=4 sw=4 et:
 """Regression interfaces."""
+import numpy as np
 import pandas as pd
 from nipype import logging
 from nipype.interfaces.base import (
@@ -10,9 +11,10 @@ from nipype.interfaces.base import (
     TraitedSpec,
     traits,
 )
+from sklearn.linear_model import LinearRegression
 
-from xcp_d.utils.filemanip import fname_presuffix
-from xcp_d.utils.utils import demean_detrend_data, linear_regression
+from xcp_d.utils.filemanip import fname_presuffix, split_filename
+from xcp_d.utils.utils import demean_detrend_data
 from xcp_d.utils.write_save import despikedatacifti, read_ndata, write_ndata
 
 LOGGER = logging.getLogger("nipype.interface")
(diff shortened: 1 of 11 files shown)


xcp_d/interfaces/regression.py
```diff
@@ -54,38 +56,60 @@ def _run_interface(self, runtime):
         # Get the confound matrix
         confound = pd.read_table(self.inputs.confounds)
 
-        confound = confound.to_numpy().T  # Transpose confounds matrix to line up with bold matrix
+        # Any columns starting with "signal__" are assumed to be signal regressors
```
tsalo, line 59: Need to document this feature in the custom confounds section.

tsalo: Done in [d1f9306](https://github.com/PennLINC/xcp_d/pull/697/commits/d1f9306ca14bf86fc10fac0dcbbcf4235adfb24a).

xcp_d/interfaces/regression.py
```diff
@@ -53,43 +54,88 @@ def _run_interface(self, runtime):
...
+            # Perform non-aggressive denoising.
+            # First, mean-center and detrend BOLD data
+            bold_arr = signal.clean(
+                signals=bold_arr,
+                detrend=True,  # this mean-centers and linearly detrends the data
+                standardize=False,
+                confounds=None,
+                filter=None,
+                ensure_finite=True,
+            )
```
tsalo, line 92: Should we also detrend the confounds?

[Per Nilearn](https://github.com/nilearn/nilearn/blob/cfcebceba4f180a7f55f9553ab1d73478961d5b8/nilearn/signal.py#L629-L631) (citing [Lindquist et al. [2019]](https://doi.org/10.1002/hbm.24528)):

>Detrend and filtering should apply to confounds, if confound[s] present
>keep filters orthogonal (according to Lindquist et al. (2018))
>Restrict the signal to the orthogonal of the confounds

tsalo: Per Matt, since we filter _after_ regression, it should be fine. While detrending and then regressing out undetrended confounds can reintroduce trends, those trends then get removed by the filtering done later.

On the other hand... users can disable bandpass filtering, so we should probably do the regression properly.

tsalo: Confounds will now be mean-center and detrended in the same manner as the BOLD data.

kahinimehta (approved):
Don't know too much about the background behind the regression methods, but the code looks good to me!

tsalo:
@smoia Thanks for taking a look!

xcp_d/interfaces/regression.py
```diff
@@ -53,43 +54,97 @@ def _run_interface(self, runtime):
...
+            noise_reg_idx = [i for i, c in enumerate(cols) if c not in signal_columns]
 
-        confound = confound.to_numpy().T  # Transpose confounds matrix to line up with bold matrix
         # Get the nifti/cifti matrix
-        bold_matrix = read_ndata(datafile=self.inputs.in_file, maskfile=self.inputs.mask)
+        bold_arr = read_ndata(datafile=self.inputs.in_file, maskfile=self.inputs.mask)
+        bold_arr = bold_arr.T  # transpose BOLD data to TxS
 
-        # Demean and detrend the data
+        # Regress out the confounds via linear regression from sklearn
```
smoia, line 73:
```suggestion
        # Regress out the confounds via linear regression from nilearn
```

smoia (commented):
To the extent of my knowledge, your implementation makes sense - what I _think_ you're trying to do is to implement `fsl_regfilt`. If that's correct, the implementation I see is righT!

xcp_d/interfaces/regression.py
tsalo: Good catch

tsalo:
@smoia that is indeed what I am trying to do. Thanks so much for reviewing this!

PR #703, opened 2022-12-05
Attempt to ingress HCP data into XCP

kahinimehta:
A fix for the code that ingresses HCP data into XCP


tsalo:
@kahinimehta can you deal with the linter errors?

I'm labeling this PR "ignore-for-release" so you can do a fancier version later.

xcp_d/utils/hcp2fmriprep.py
```diff
@@ -10,18 +10,47 @@
...
 
 from xcp_d.utils.dcan2fmriprep import copyfileobj_example, extractreg, writejson
+import logging
+
+LOGGER = logging.getLogger("hcp")
 
 
 def hcp2fmriprep(hcpdir, outdir, sub_id=None):
     """Convert HCP-format data to fMRIPrep format."""
+    LOGGER.warning(f"This is an experimental function and has not been tested yet.")
```
tsalo, line 20:
```suggestion
    LOGGER.warning("This is an experimental function and has not been tested yet.")
```

xcp_d/utils/bids.py
```diff
@@ -152,12 +152,21 @@ def collect_data(
...
-        derivatives=True,
-        config=["bids", "derivatives"],
-    )
+    if input_type != 'hcp':
+        layout = BIDSLayout(
+            str(bids_dir),
+            validate=bids_validate,
+            derivatives=True,
+            config=["bids", "derivatives"],
+        )
```
tsalo, line 161: I believe that just about everything that applies to "hcp" data will apply to "dcan" data as well.

xcp_d/utils/bids.py
```diff
@@ -152,12 +152,20 @@ def collect_data(
     layout : pybids.layout.BIDSLayout
     subj_data : dict
     """
-    layout = BIDSLayout(
-        str(bids_dir),
-        validate=bids_validate,
-        derivatives=True,
-        config=["bids", "derivatives"],
-    )
+    if input_type != "hcp" or input_type != "dcan":
```
tsalo, line 155:
```suggestion
    if input_type not in ("dcan", "hcp"):
```

xcp_d/utils/bids.py
```diff
@@ -152,12 +152,20 @@ def collect_data(
...
+        layout = BIDSLayout(
+            str(bids_dir),
+            validate=bids_validate,
+            derivatives=True,
+            config=["bids", "derivatives"],
+        )
+    else:
+        layout = BIDSLayout(
```
tsalo, line 163: Can you add a comment here why you set `derivatives=False` for dcan and hcp data?

xcp_d/utils/bids.py
```diff
@@ -264,6 +272,9 @@ def collect_data(
         )["nifti"]
         for space in temp_allowed_spaces:
             temp_bold_query["space"] = space
+            if input_type == "hcp" or input_type == "dcan":
```
tsalo, line 275:
```suggestion
            if input_type in ("hcp", "dcan"):
```

tsalo: Can you add a comment about why this is done? I assume it's because there's no preprocessed BOLD data in nifti format, and you're using the boldref instead?

Also, I think you can move this if statement to outside the for loop, since it's looping through spaces, and `desc` and `suffix` should stay the same across space-specific queries.

xcp_d/utils/bids.py
```diff
@@ -287,6 +298,10 @@ def collect_data(
     if len(densities) > 1:
         queries["bold"]["density"] = densities[0]
 
+    if input_type == "hcp" or input_type == "dcan":
+        queries["template_to_t1w_xform"]["from"] = "MNI152NLin2009cAsym"
+        queries["t1w_to_template_xform"]["to"] = "MNI152NLin2009cAsym"
```
tsalo, line 303: If you want to hardcode the space in, then I would recommend only looping through spaces above if `input_type` isn't hcp or dcan.

xcp_d/utils/bids.py
```diff
@@ -341,7 +358,7 @@ def collect_run_data(layout, bold_file, cifti=False):
     if "RepetitionTime" not in metadata["bold_metadata"].keys():
         metadata["bold_metadata"]["RepetitionTime"] = _get_tr(bold_file)
 
-    if not cifti:
+    if not cifti and (input_type != "hcp" or input_type != "dcan"):
```
tsalo, line 361:
```suggestion
    if not cifti and input_type not in ("hcp", "dcan"):
```

xcp_d/utils/bids.py
```diff
@@ -361,6 +378,23 @@ def collect_run_data(layout, bold_file, cifti=False):
             suffix="xfm",
         )
 
+    if not cifti and (input_type == "hcp" or input_type == "dcan"):
```
tsalo, line 381:
```suggestion
    elif not cifti:
```

xcp_d/utils/bids.py
```diff
@@ -441,9 +475,9 @@ def get_preproc_pipeline_info(input_type, fmri_dir):
     if os.path.isfile(dataset_description):
         with open(dataset_description) as f:
             dataset_dict = json.load(f)
-
+    try:
         info_dict["version"] = dataset_dict["GeneratedBy"][0]["Version"]
-    else:
+    except:
```
tsalo, line 480: I try to avoid try/except statements if possible. What does the HCP dataset description file look like?

xcp_d/utils/hcp2fmriprep.py
```diff
@@ -113,30 +143,61 @@ def hcpfmriprepx(hcp_dir, out_dir, subid):
...
         # convert rot to rad
         mvreg["rot_x"] = mvreg["rot_x"] * np.pi / 180
         mvreg["rot_y"] = mvreg["rot_y"] * np.pi / 180
         mvreg["rot_z"] = mvreg["rot_z"] * np.pi / 180
+        mvreg["rot_x_derivative1"] = mvreg["rot_x_derivative1"] * np.pi / 180
+        mvreg["rot_y_derivative1"] = mvreg["rot_y_derivative1"] * np.pi / 180
+        mvreg["rot_z_derivative1"] = mvreg["rot_z_derivative1"] * np.pi / 180
+        mvreg["trans_x_power2"] = mvreg["trans_x"] ** 2
+
+        # get derivatives and powers
```
tsalo, line 189:
```suggestion

        # get derivatives and powers
        mvreg["trans_x_power2"] = mvreg["trans_x"] ** 2
```
Just moving it down to be with similar assignments.

xcp_d/workflow/anatomical.py
```diff
@@ -316,6 +316,18 @@ def init_anatomical_wf(
         for ss in surf:
             shutil.copy(ss, anatdir)
 
+        # fmt:off
+        nothingnode = pe.Node(
+            niu.IdentityInterface(fields=["t1w", "t1seg"]),
+            name="nothingnode",
+        )
```
tsalo, line 323:
```suggestion
        nothingnode = pe.Node(
            niu.IdentityInterface(fields=["t1w", "t1seg"]),
            name="nothingnode",
        )
        # fmt:off
```

xcp_d/workflow/anatomical.py
```diff
@@ -316,6 +316,18 @@ def init_anatomical_wf(
...
+        # fmt:off
+        nothingnode = pe.Node(
+            niu.IdentityInterface(fields=["t1w", "t1seg"]),
+            name="nothingnode",
+        )
+        workflow.connect(
+            [(inputnode, nothingnode, [("t1w", "t1w")]),
+             (inputnode, nothingnode, [("t1seg", "t1seg")]),
+             ]
+        )
```
tsalo, line 328:
```suggestion
        workflow.connect([
            (inputnode, nothingnode, [
                ("t1w", "t1w"),
                ("t1seg", "t1seg"),
            ]),
        ])

```

xcp_d/workflow/cifti.py
```diff
@@ -91,13 +92,16 @@ def init_ciftipostprocess_wf(
                 despike=False,
                 dcan_qc=False,
                 n_runs=1,
+                input_type,
                 layout=None,
+                input_type,
```
tsalo, line 97: `input_type` is in here twice. Also, without feeding it in as a keyword argument (e.g., `input_type="fmriprep"`), it would fail, since it comes after keyword arguments (e.g., `n_runs=1`).

xcp_d/workflow/cifti.py
```diff
@@ -125,6 +129,7 @@ def init_ciftipostprocess_wf(
         BIDS dataset layout
     %(name)s
         Default is 'cifti_postprocess_wf'.
+    input_type
```
tsalo, line 132:
```suggestion
```

It's already in there, after bold_file.

xcp_d/workflow/bold.py
```diff
@@ -152,12 +153,14 @@ def init_boldpostprocess_wf(
         Fed from the subject workflow.
     fmriprep_confounds_tsv
         Loaded in this workflow.
+    input_type
+        Input type.
```
tsalo, line 157: This should be listed under parameters, not inputs.

tsalo (changes requested):
Thanks for this. I have a few minor requested changes.

xcp_d/utils/bids.py
tsalo: Your comment says that the files are already written into a derivatives directory, but that's true for fMRIPrep as well. The preprocessed data, whether it comes from HCP, DCAN, fMRIPrep, or Nibabies, is in a BIDS derivatives dataset.

xcp_d/utils/bids.py
tsalo: Sorry, why did you resolve this?

kahinimehta: I think it doesn't loop through for temp_queries anymore

xcp_d/utils/bids.py
```diff
@@ -262,13 +271,27 @@ def collect_data(
...
+            # nifti data, we will use the boldref
+            temp_bold_query["desc"] = None
+            temp_bold_query["suffix"] = "boldref"
+            temp_bold_query["space"] = "MNI152NLin6Asym"
+
+        nifti_bold_data = layout.get(**temp_bold_query)
+
+        if nifti_bold_data:
+            queries["t1w_to_template_xform"]["to"] = space
+            queries["template_to_t1w_xform"]["from"] = space
```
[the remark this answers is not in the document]
kahinimehta: It seems the boldref is in NLin6, the T1 is in 2009c?

xcp_d/utils/bids.py
tsalo: Does it have a "GeneratedBy" field?

xcp_d/utils/bids.py
```diff
@@ -262,13 +271,25 @@ def collect_data(
...
+
+        else:  # HCP and DCAN files don't have
+            # nifti data, we will use the boldref
+            temp_bold_query["desc"] = None
+            temp_bold_query["suffix"] = "boldref"
+            temp_bold_query["space"] = "MNI152NLin6Asym"
+            queries["template_to_t1w_xform"]["from"] = "MNI152NLin2009cAsym"
+            queries["t1w_to_template_xform"]["to"] = "MNI152NLin2009cAsym"
+
+        nifti_bold_data = layout.get(**temp_bold_query)
```
tsalo, line 292:
```suggestion
        if input_type in ("hcp", "dcan"):
            temp_allowed_spaces = ["MNI152NLin2009cAsym"]
            # HCP and DCAN files don't have nifti BOLD data, we will use the boldref
            temp_bold_query["desc"] = None
            temp_bold_query["suffix"] = "boldref"

        for space in temp_allowed_spaces:
            temp_bold_query["space"] = space
            nifti_bold_data = layout.get(**temp_bold_query)
            if nifti_bold_data:
                queries["t1w_to_template_xform"]["to"] = space
                queries["template_to_t1w_xform"]["from"] = space
                break
```

xcp_d/utils/bids.py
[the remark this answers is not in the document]
tsalo: I think that may be a problem.

So, we have standard-space BOLD data in MNI152NLin6Asym, but standard-space T1w images in MNI152NLin2009cAsym. That BOLD data is only used to identify the template to use in the two xforms. There's no reason to query for a BOLD file in one space, but transforms in another.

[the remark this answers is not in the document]
kahinimehta: Resolved during scrum: it shouldn't technically matter, but we will want to refactor this code later.

xcp_d/utils/bids.py
kahinimehta: Yes!

xcp_d/utils/bids.py
kahinimehta: Yes

xcp_d/utils/bids.py
```diff
@@ -270,6 +270,16 @@ def collect_data(
...
 
+        if input_type in ("hcp", "dcan"):
+            temp_allowed_spaces = ["MNI152NLin6Asym"]
+            # HCP and DCAN files don't have nifti BOLD data, we will use the boldref
+            temp_bold_query["desc"] = None
+            temp_bold_query["suffix"] = "boldref"
+            temp_bold_query["space"] = "MNI152NLin6Asym"
+            queries["t1w_to_template_xform"]["to"] = "MNI152NLin2009cAsym"
+            queries["template_to_t1w_xform"]["from"] = "MNI152NLin2009cAsym"
+            nifti_bold_data = layout.get(**temp_bold_query)
```
tsalo, line 281: Just want to make a note that this is a hack that we should drop. I'm thinking we should probably search directly for transforms, since the BOLD file isn't actually used. That can go in a future PR though.

xcp_d/utils/bids.py
```diff
@@ -361,6 +373,21 @@ def collect_run_data(layout, bold_file, cifti=False):
...
+            suffix="boldref",
+        )
+        run_data["boldmask"] = layout.get(
+            return_type="file",
+            suffix="mask",
+            desc="brain",
+        )
+        run_data["t1w_to_native_xform"] = layout.get(
+            return_type="file", datatype="anat", suffix="xfm", to="MNI152NLin2009cAsym"
+        )
```
tsalo, line 389: Just noticed this. It looks like the T1w-to-native transform is set to be a transform from any anatomical space (which probably would just be the T1w space) to _MNI_ space, rather than native space.

tsalo: I'm guessing this isn't a real problem in practice, since the HCP "transform files" are just copies of its preprocessed T1w file, which is a separate thing we need to deal with.

xcp_d/utils/hcp2fmriprep.py
```diff
@@ -149,14 +210,34 @@ def hcpfmriprepx(hcp_dir, out_dir, subid):
...
+
+        regressors["white_matter_derivative1"] = pd.DataFrame(
+            np.diff(regressors["white_matter"].to_numpy(), prepend=0)
+        )
+        regressors["white_matter_derivative1_power2"] = regressors["white_matter_derivative1"] ** 2
+        regressors["white_matter_power2"] = regressors["white_matter"] ** 2
+        regressors["csf_power2"] = regressors["csf"] ** 2
+        regressors["csf_derivative1"] = pd.DataFrame(
+            np.diff(regressors["csf"].to_numpy(), prepend=0)
+        )
```
tsalo, line 232: I think `load_confounds` prepends the first value, rather than zero. Worth looking at later.

xcp_d/utils/hcp2fmriprep.py
```diff
@@ -113,30 +143,61 @@ def hcpfmriprepx(hcp_dir, out_dir, subid):
...
+    print(hcp_dir)
     for j in taskx:
         if j.endswith("RL") or j.endswith("LR"):
             tasklistx.append(j)
 
-    csf_mask = pkgrf("xcp_d", "masks/csf.nii.gz")
-    wm_mask = pkgrf("xcp_d", "masks/wm.nii.gz")
-
+    csf_mask = pkgrf("xcp_d", "/data/masks/csf.nii.gz")
+    wm_mask = pkgrf("xcp_d", "/data/masks/wm.nii.gz")
```
tsalo, line 157: We should rename these to be more informative. I'm guessing they're in a specific MNI space+resolution.

xcp_d/utils/hcp2fmriprep.py
```diff
@@ -113,30 +143,61 @@ def hcpfmriprepx(hcp_dir, out_dir, subid):
...
     for k in tasklistx:
         idx = os.path.basename(k).split("_")
         filenamex = os.path.basename(k)
 
         # create confound regressors
 
         mvreg = pd.read_csv(k + "/Movement_Regressors.txt", header=None, delimiter=r"\s+")
-        mvreg = mvreg.iloc[:, 0:6]
-        mvreg.columns = ["trans_x", "trans_y", "trans_z", "rot_x", "rot_y", "rot_z"]
+        mvreg = mvreg.iloc[:, :]
```
tsalo, line 165: This is probably unnecessary.

tsalo (approved):
Thanks for this. I think it's good to merge now, but I want to make sure we open issues regarding all of the problems we've noticed. There are some bugs in the code that will only reveal themselves when we've fixed the actual HCP ingression, but I've tried to note them where I could.

xcp_d/utils/bids.py
kahinimehta: Will make a note

xcp_d/utils/hcp2fmriprep.py
kahinimehta: I just copied the code from what XCP was using previously, but will make a note!

xcp_d/utils/hcp2fmriprep.py
kahinimehta: will make a note

xcp_d/utils/hcp2fmriprep.py
kahinimehta: Ahh yes, this was updated later

xcp_d/utils/bids.py
tsalo, line 389: @kahinimehta do you know why you had to add this part? I just realized that running xcp_d on HCP or DCAN data should automatically enable cifti processing, so AFAICT this shouldn't be reachable.

kahinimehta: I think this was supposed to address the "input_type not in ("hcp", "dcan)""  part of the loop

tsalo: This should catch when input_type is either "hcp" or "dcan", but is there any situation when we'd run nifti processing on hcp or dcan data?

Issue #704, opened 2022-12-07
Use native-space surface files in brainsprite when standard-space ones are unavailable


kahinimehta:
aparc+aseg files are used as 'dseg' in the HCP and DCAN workflows - but we are not sure if these are appropriate to use in the fMRIPrep and Nibabies workflows. We need to make sure that the brainsprite works with an aparc+aseg file.

ericfeczko:
This doesn't make any sense to me -- the brainsprite viewer should be working directly to visualize the surfaces themselves on top of the anatomical images -- there should be no reason to use a ribbon or segmentation file. 

In fact, I recommend against it as it will lead to inaccurate viewing of the surfaces on the volumes.

tsalo:
The updated plan (thanks @ericfeczko) is to have three possible routes:

1. Native-space surface files are available, and the user has included the `--warp-surfaces-native2std` flag. In this scenario, we'll take the standard-space surfaces from the anatomical workflow (post-#706) and plot those on the standard-space T1w image.
2. Native-space surface files are available, but the user has **not** included the `--warp-surfaces-native2std` flag. In this case, we'll take the native-space surfaces and plot those on top of the native-space T1w image.
3. Surface files are not available (e.g., the user ran fMRIPrep with `--fs-no-reconall`). In this situation, we just won't make the brainsprite.

tsalo:
I'm not sure if the scene files will work with native-space surfaces/T1w. We'll have to see.

ericfeczko:
Interesting -- I'd check with Robert Hermosillo; workbench shouldn't have any issues with loading native surfaces though.

PR #702, opened 2022-12-02
Adopt executive summary's brainsprite using jinja templates

tsalo:
Closes #645 and closes #704.

## Changes proposed in this pull request
- Replace nilearn brainsprite with Connectome Workbench's show-scene-based mosaic, followed by custom brainsprite code from [`ExecutiveSummary`](https://github.com/DCAN-Labs/ExecutiveSummary).
- Use jinja to build the executive summary.
- Drop the brainsprite from the nipreps-based report.
- Create new function, `xcp_d.utils.bids.get_entity()`, which extracts a given entity from a filename using regular expressions.
- Remove `xcp_d.interfaces.report.get_space()` and `xcp_d.utils.bids._getsesid()`.

PR #706, opened 2022-12-08
Refactor anatomical workflow

tsalo:
Closes None, but prepares for #704.

- Stop outputting fMRIPrep-style midthickness files. **This is the breaking change.**
- Separate the anatomical workflow into four:
    - `init_warp_surfaces_to_template_wf `: (Renamed from `init_anatomical_wf`) Wrangle the surfaces, call the other two workflows, and calculate HCP-style surfaces if necessary.
    - `init_update_xform_wf`: Modify fMRIPrep-generated T1w-to-template transforms so that they can be applied with wb_connect to surface files to get them into fsLR space.
        - Does this mean the template in question should **always** be MNI152NLin6Asym?
    - `init_warp_one_hemisphere_wf`: Apply the modified transforms to warp native-space surface files to standard space.
    - `init_inflate_surfaces_wf`: Inflate midthickness file and save out either inflated+very-inflated or just very-inflated, depending on parameter. This way, we can write out the very-inflated HCP-style file for DCAN/HCP derivatives.
- Use MapNodes in the anatomical workflow, when possible.
    - I think we could probably nest the iteration across both hemisphere and surface, but I'm not 100% sure how to do that.
- Return standard-space surfaces from anatomical workflow and feed them into the brainsprite workflow. These are currently unused by the brainsprite workflow, but will eventually work with #702.
- ~~Collect surfaces in a new function, using BIDSLayout, instead of globbing for them within the anatomical workflow.~~ Moved to #731.
    - Expect pial and smoothwm surface files from fMRIPrep and Nibabies derivatives.
    - Expect HCP-style midthickness, inflated, and very-inflated files from DCAN and HCP derivatives.
- ~~Move FreeSurfer folder detection to its own function.~~ Moved to #719.

The workflow.anatomical module's workflows' API pages would be good to review. They now have workflow graphs!


xcp_d/workflow/anatomical.py
```diff
@@ -269,1218 +297,1136 @@ def init_anatomical_wf(
...
+            n_procs=omp_nthreads,
+            iterfield=["in_file"],
+        )
+        rh_native_apply_warpfield = pe.MapNode(
+            ApplyWarpfield(),
+            name="rh_native_apply_warpfield",
+            mem_gb=mem_gb,
+            n_procs=omp_nthreads,
+            iterfield=["in_file"],
+        )
```
tsalo, line 576: @madisoth do you know why these nodes do not have outgoing connections? Is this branch of the workflow unused?

xcp_d/workflow/anatomical.py
```diff
@@ -269,1218 +297,1136 @@ def init_anatomical_wf(
...
-                    ),
-                ]
-            )
+        # make "HCP-style" native midthickness and inflated
+        lh_native_hcpmidthick_wf = pe.Node(
+            SurfaceAverage(),
+            name="lh_native_hcpmidthick_wf",
+            mem_gb=mem_gb,
+            n_procs=omp_nthreads,
+        )
```
tsalo, line 766: @madisoth AFAICT, when standard-space surfaces are available (e.g., with the DCAN or HCP pipelines), the HCP-style files will be missing. My assumption is that those pipelines create _just_ the HCP-style files, and they are just mislabeled in XCPD. Is that correct?

If not, should we be creating the HCP-style surface files when standard-space surfaces are already available?

xcp_d/workflow/anatomical.py
```diff
@@ -269,1218 +297,1136 @@ def init_anatomical_wf(
...
-            L_pial_surf,
-            R_pial_surf,
-            L_wm_surf,
-            R_wm_surf,
-        ]
+        # fmt:off
+        workflow.connect([
+            (inputnode, merge_files_to_list, [
+                ("lh_inflated_surf", "in1"),
+                ("rh_inflated_surf", "in2"),
```
tsalo, line 345: @madisoth AFAICT, the inflated surfaces aren't warped to standard space, and aren't used for anything except naming HCP-style output files generated later in the workflow. Is there a reason these files are included?

xcp_d/workflow/anatomical.py
```diff
@@ -269,1218 +261,932 @@ def init_anatomical_wf(
...
+        ApplyTransforms(
+            reference_image=mnitemplate,
+            interpolation="LanczosWindowedSinc",
+            print_out_composite_warp_file=True,
+            output_image="ants_composite_inv_xfm.nii.gz",
+        ),
+        name="combine_inv_xfms",
+        mem_gb=mem_gb,
+        n_procs=omp_nthreads,
+    )
```
tsalo, line 785: @madisoth, both of these nodes take inputs, but don't feed out to anything in the workflow. Is there a reason for this?

xcp_d/workflow/anatomical.py
madisoth: They're unused and can be removed. 

At one point we planned to output surfaces at 3 densities as in HCP (native, 164k, 32k) which this is left over from

xcp_d/workflow/anatomical.py
madisoth: Yes, that's correct

xcp_d/workflow/anatomical.py
madisoth: Correct, they're only used for naming the HCP-style outputs

Originally we did warp them to standard space but it often fail due to the surfaces extending outside the FOV of the warp.

tsalo: Ah, I see. In that case, we don't really need them at all, since we know how the HCP-style inflated files should be named based on other surfaces. I will remove the regular inflated files entirely.

xcp_d/workflow/anatomical.py
tsalo: In that case, it looks like I need to (1) ensure the HCP-style surfaces coming out of the DCAN and HCP conversion functions include `desc-hcp`, (2) check whether datasets have `desc-hcp` or plain standard-space surfaces, and (3) toggle HCP-style surface creation based on that. Thanks for the clarification.

xcp_d/workflow/anatomical.py
madisoth: At one point I planned to use the combined warps on the surfaces when possible (i.e. when the surfaces fit within the FOV of the warpfield), and fall through to using the separate affine and warpfield xfms otherwise); feel free to remove this

tsalo: Thanks. I'll remove them then.

xcp_d/workflow/anatomical.py
tsalo: Well, we do need to write out the inflated files to the derivatives folder _if_ they're already in standard space and are (presumably) HCP-style, so I will limit them instead of removing them.

xcp_d/workflow/anatomical.py
tsalo: I think I need to dig into this a bit more.

1. Is there any situation where the preprocessing workflow might produce fsLR-space midthickness and [very-]inflated files that _aren't_ in the HCP style?
2. If the answer to the above is "yes", would it be better to use the native-space surface files to make HCP-style fsLR-space surfaces (i.e., run the whole workflow anyway), or just output the non-HCP-style fsLR-space surfaces and call it a day?
3. If we have HCP-style midthickness and inflated files, should we also create the very-inflated files? Currently, the workflow doesn't do this, but it seems pretty straightforward.
4. Do we really need to output the non-HCP-style midthickness file? This seems to be the only file that wouldn't match between the two branches of the tree. It gets written out if there are only native-space surfaces, but doesn't if there are standard-space (presumably HCP-style) surfaces.

xcp_d/workflow/anatomical.py
```diff
@@ -255,8 +248,27 @@ def init_anatomical_wf(
     ------
     t1w : str
         Path to the T1w file.
-    t1seg : str
-        Path to the T1w segmentation file.
+        Only seems to be used for a stray branch of init_update_xform_wf.
+    t1w_to_template_xform
+    template_to_t1w_xform
```
tsalo, line 253: @madisoth our assumption is that the selected volumetric template from `collect_data` is applicable here. However, since the target space is fsLR, which uses MNI152NLin6Asym for its volumetric data, does the template in question _always_ have to be MNI152NLin6Asym?

xcp_d/workflow/anatomical.py
```diff
@@ -229,1272 +222,886 @@ def init_anatomical_wf(
...
     -----
     If "hcp" or "dcan" input type, pre-generated surface files will be collected from the
     converted preprocessed derivatives.
     However, these derivatives do not include HCP-style surfaces.
+    Unless maybe they *are* HCP-style surfaces already?
 
-    If "fmriprep" or "nibabies", surface files in fsnative space will be extracted from the
-    associated Freesurfer directory (if available), and warped to fsLR space.
+    If "fmriprep" or "nibabies", surface files in fsnative space will be extracted from
+    the associated Freesurfer directory (if available), and warped to fsLR space.
```
tsalo, line 278: @madisoth the fMRIPrep anatomical derivatives include a T1w-to-fsnative transform. Just wondering, what is the purpose of that transform, and does it need to be incorporated in the anatomical workflow?

xcp_d/workflow/anatomical.py
madisoth: 1. Currently no; this would only change if fsLR surface outputs are added to fMRIPrep / NiBabies in the future.
3. DCAN/HCP derivatives should include all 3 (midthickness, inflated, vinflated) in fsLR32k but if inflated/vinflated are missing then it would be good to generate them
4. I'm okay with omitting non-HCP midthickness.

xcp_d/workflow/anatomical.py
madisoth: That's correct for all 4 of DCAN-HCP / HCP / fMRIPrep / NiBabies.


Supporting other, nonstandard volume spaces for CIFTI could help enable ingress from things like the DCAN macaque pipeline (HCP-based pipeline which outputs fsLR surfaces but non-MNI152 volumes) in the future I suppose.

xcp_d/workflow/anatomical.py
madisoth: It's the affine transform from subject T1w to FreeSurfer volume space ("fsnative" perhaps a misnomer as that's the surface space). The anatomical workflow doesn't need it (nor the inverse xfm)

madisoth (approved):

xcp_d/workflow/anatomical.py
tsalo: In that case, I think I need to anchor each CIFTI space we support to a specific volumetric space. At the moment, that would just be fsLR --> MNI152NLin6Asym, but I can try to write the code so it can be expanded easily.

xcp_d/workflow/anatomical.py
tsalo: Thanks!

xcp_d/workflow/anatomical.py
tsalo: >2. DCAN/HCP derivatives should include all 3 (midthickness, inflated, vinflated) in fsLR32k but if inflated/vinflated are missing then it would be good to generate them

The conversion functions don't include vinflated files. Would you mind taking a look at the following and letting me know what the vinflated files would be named?

https://github.com/PennLINC/xcp_d/blob/57fa13f7712f4a552b58202dd83f5b18ca6d7339/xcp_d/utils/hcp2fmriprep.py#L78-L87

xcp_d/workflow/anatomical.py
tsalo: I've opened #720 for this.

tsalo:
@madisoth would you mind reviewing the changes in a52ab05?

Here's a summary of the changes:

- No longer generate or write out fMRIPrep-style midthickness files. Only HCP-style midthickness, inflated, and very-inflated files from now on.
- Try to find HCP-style midthickness, inflated, and very-inflated surface files in `collect_surface_data`. If they aren't found (and the pial/wm files are in native space), then the anatomical workflows will make them.
- Rename `init_anatomical_wf` to `init_warp_surfaces_to_template_wf`.
- Expect very-inflated surface files from the HCP and DCAN pipelines. This ties in with https://github.com/PennLINC/xcp_d/pull/714#discussion_r1047414712.

xcp_d/workflow/anatomical.py
tsalo: Resolved in #714.

tsalo:
In the DCAN pipeline, the HCP-style midthickness file is generated _after_ warping to fsLR-32k, so instead of inflating before warping, we can calculate the midthickness and then inflate, all after warping.

tsalo:
Just tackled standard-space midthickness in [d3cf9cf](https://github.com/PennLINC/xcp_d/pull/706/commits/d3cf9cfe025dd9124d4df6226f2cc9a06502b5ec).

madisoth (approved):

PR #720, opened 2022-12-13
Only use associated vol space for ciftis

tsalo:
Closes None, but stems from https://github.com/PennLINC/xcp_d/pull/706#discussion_r1047209856.

- When processing CIFTI data and looking for a volumetric space, only look for the volumetric space associated with the CIFTI data's template. At the moment, this means fsLR --> MNI152NLin6Asym, but it could be expanded in the future.

None.


tsalo:
Something I should have realized is that there's no guarantee that the appropriate volumetric data will be present.

PR #714, opened 2022-12-09
Refactor HCP/DCAN ingression and fix converted filenames

tsalo:
Related to #709.

To do:

- [x] Test on real data.
- [ ] Figure out how to support nifti transform files. I assume that any tools that use the transforms can handle either file type, and the issue is just in collecting the appropriate files.

## Changes proposed in this pull request
- Replace string concatenation with os.path.join calls.
- Keep track of input and output files with a dictionary, and write that dictionary out to the output directory.
- Improve variable names.
    - Just as an example, `midR` was used for the HCP right-hemisphere midthickness file, while ``rMid` was used for the fMRIPrep version.

## Documentation that should be reviewed
None.

PR #721, opened 2022-12-14
Refactor executive summary workflow

tsalo:
Closes None.

Open questions:
- When would we want to warp the T1w image to standard space? At the moment, we collect the native T1w-space image, but we should really collect the standard-space one as well and just save that to the output directory. When would there be BOLD data in standard space, but not a T1w image in the same space?
- We don't support native or T1w-space inputs/outputs, so do we need to support warps to T1w or native space?

Resolved questions:
- Should we just use the dseg in the preprocessing derivatives?
    - There should be a standard-space tissue-type dseg in the fMRIPrep derivatives, in the anat folder. **However, this dseg is just the three tissue types, I believe, while the carpet plot uses labels like subcortical GM and cerebellum.**
    - I assume that Nibabies derivatives would include the same thing.
    - I'm not sure about the DCAN and HCP derivatives. In #714, it became clear that what we called the regular dseg was actually the aparc+aseg file, so we may not have a tissue type dseg file from those pipelines. On the other hand, would we ever run DCAN/HCP with NIFTI processing (i.e., not CIFTI processing)? If the answer is "no", then we won't ever _need_ a tissue type dseg from those pipelines.
        - The answer _is_ "no", since processing HCP/DCAN data will automatically set cifti to True.

- Move the carpet plot generation code from the executive summary workflow into the QC plotting workflow.
    - This streamlines things since the same segmentation file is used for both sets of carpet plots and the whole warp-to-BOLD-space string of Nodes are the same between the two.
- Use the T1w-to-native transform from `collect_run_data` instead of trying to find it with a custom function.
- Load a boldref file in `collect_run_data`, even if running the CIFTI workflow, which can then be plotted in the executive summary workflow. This replaces the function that tried to find a boldref file.
- Remove the HCP/DCAN hack in `collect_data`. Just search for the MNI/T1w transforms directly, instead of using NIFTI BOLD files as an intermediate.
- Remove unused arguments from the executive summary workflow, now that it only makes the boldref and registration plots.
- ~~Fix a bug in the NIFTI executive summary, where the segmentation file used for the carpet plots was in MNI152NLin2009cAsym space, but transforms from MNI152NLin6Asym were used. I've tacked on the MNI152NLin2009cAsym --> MNI152NLin6Asym transform, to get the dseg from MNI152NLin2009cAsym space to the BOLD space.~~ Moved to #727.

### DIFF
--- a/xcp_d/interfaces/surfplotting.py
+++ b/xcp_d/interfaces/surfplotting.py
@@ -8,6 +8,7 @@ from nipype.interfaces.base import (
     File,
     SimpleInterface,
     TraitedSpec,
+    isdefined,
     traits,
 )
 
@@ -130,6 +131,12 @@ class PlotSVGData(SimpleInterface):
             use_ext=False,
         )
 
+        mask_file = self.inputs.mask
+        mask_file = mask_file if isdefined(mask_file) else None
+
+        segmentation_file = self.inputs.seg_data
+        segmentation_file = segmentation_file if isdefined(segmentation_file) else None
+
         self._results["before_process"], self._results["after_process"] = plot_svgx(
             preprocessed_file=self.inputs.rawdata,
             residuals_file=self.inputs.regressed_data,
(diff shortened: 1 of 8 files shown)


tsalo:
I think I might just grab any old standard-space boldref in collect_run_data, and not rely on template_to_t1w for CIFTIs at all. This way, we don't need to care about volumetric spaces with CIFTIs.

tsalo:
>I'm not sure about the DCAN and HCP derivatives. In https://github.com/PennLINC/xcp_d/pull/714, it became clear that what we called the regular dseg was actually the aparc+aseg file, so we may not have a tissue type dseg file from those pipelines. On the other hand, would we ever run DCAN/HCP with NIFTI processing (i.e., not CIFTI processing)? If the answer is "no", then we won't ever need a tissue type dseg from those pipelines.

@madisoth would you mind weighing in on this?

PR #742, opened 2023-01-03
Remove patched interfaces

tsalo:
Closes None.

- Remove DespikePatch and hardcode the output filename.
- Remove ReHoNamePatch and hardcode the output filename.

None.

### DIFF
--- a/.circleci/RunPyTests.sh
+++ b/.circleci/RunPyTests.sh
@@ -39,7 +39,7 @@ run_pytest_cmd () {
     cwd_mount="-v ${PWD}/tests:/rundir:rw"
     PYTEST_RUN="docker run --rm -ti -u $(id -u) -w /rundir --entrypoint pytest "
     PYTEST_RUN+="${workdir_mount} ${patch_mount} ${cwd_mount} ${cfg_arg} ${bids_mount} ${output_mount} ${IMAGE} "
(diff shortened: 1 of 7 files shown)


tsalo:
It looks like the ReHo patch interface copies the input file, but the Despike patch doesn't. I can adapt one or more of the upstream nodes to generate shorter output filenames instead of appending to the input filenames, which should resolve the problem without requiring additional custom interfaces.

mattcieslak:
Do you know if the directories are included in the path given as input ReHo? The directories get extremely nested sometimes, which contributed to the problem. I don't think the upstream nodes can write outside of their directories.

Maybe using CopyFile in the input spec's trait for the in file will get around the problem? I'm not sure if that still ends up including the absolute path to the input file even if it's symlinked into the sandbox directory

tsalo:
I believe the absolute path is included in the filename, but it looks like our current patch interface doesn't account for that either. It just uses the current working directory, which should include the long path to the node's assigned working directory. I think the interface _could have_ used the basename of the copied file, since that file would be in the working directory, but it doesn't.

mattcieslak:
The bug appeared when there are more than 256 characters in the file name. I'd test with a 1000 character filename

tsalo:
It looks like Python has an issue with filenames longer than 256 characters, since I was getting OSErrors. I'm thinking that that means that Python is either more restrictive than, or equally restrictive as, AFNI.

tsalo:
Paused until there's a solid reason to move forward.

PR #743, opened 2023-01-05
Replace missing vertices' values with NaNs

### LINKED ISSUE #700: Parcel-wise coverage checking
## Summary
In prior versions of XCP, we checked coverage of each parcel in an atlas vs. the functional mask; if it was <50% coverage the timeseries was given NAs (and then reflected in downstream connectivity matricies).  I am not sure how this is handled now in the nifti workflow.  Separately, I am unsure how ciftis handles this issue in general (incomplete coverage).  This tends to be a big issue in large datasets with tight FOVs where coverage of dorsal/ventral brain regions (esp cerebellum, temporal horn, vertex) is often missing. 

## Next steps
a) review how the nifti workflow handles this and consider re-implementing this feature if not present
b) learn how ciftis handle incomplete coverage.

tsalo:
Closes #700. From what I can tell, partial coverage will result in time series of all zeros in the preprocessed CIFTI files- at least in fMRIPrep. I don't think CiftiParcellate ignores vertices with values that are all zeros, so those zeros probably end up being used in the averaging process. However, there is an option to ignore non-numeric values, like NaNs, so if we replace all-zero time series with all-NaN time series, those missing values should be ignored.

- Add `only_numeric` option to `CiftiParcellate`. This will ignore any non-numeric vertices (e.g., NaNs) from the summarization measure.
- Create new interface, `CiftiZerosToNaNs`, which will identify any vertices with all zeros or all NaNs and replace those vertices' time series with an array of NaNs.
- Patch the new interface into the CIFTI connectivity workflow.

None.


tsalo:
I still need to write a test or two to explicitly check this.

tsalo:
I think my test is failing because the atlas and data files have different affines or something. I assumed that CIFTIs didn't have this issue when they're in the same space, but I can't figure out how there'd be a mismatch otherwise.

PR #754, opened 2023-01-17
Add option to estimate brain radius from brain mask

### LINKED ISSUE #298: Estimate head/brain size automatically from brain mask
Scale brain size flag based off of the age of the participant. Will need to find or make a table for the best brain size for each age.

We talked about this on slack briefly - adding it here as an issue.

I think we should be able to help find / make a table. We should have a resource we can use through the BCP data.

tsalo:
Closes #298.

- Add "auto" option for `--head-radius` parameter. This option estimates the brain radius from the preprocessed brain mask by assuming the brain is a sphere, per Paul Taylor's suggestion in https://neurostars.org/t/estimating-head-brain-radius-automatically/24290/2.
- Test the new option with a unit test and one of the integration tests.

None.


tsalo:
I want to add the head radius as metadata to the filtered motion file's JSON file.

tsalo:
The metadata have been added.

PR #770, opened 2023-01-24
Require the `--cifti` flag in order to use `--warp-surfaces-native2std`

### LINKED ISSUE #765: Require CIFTI processing in order to run `--warp-surfaces-native2std`
## Summary
Allowing `--warp-surfaces-native2std` with NIFTI processing makes things more complicated, so I think it makes more sense to just block that combination.

I'm happy to discuss this if anyone really wants NIFTI processing with fsLR surfaces.

tsalo:
Closes #765.

- Only enable the surface-warping workflow if the cifti flag is also enabled. Otherwise, raise an error.
- Reorganize the CLI help strings a bit.

None?


tsalo:
I don't know why `nifti_with_freesurfer` is running all the way through. @mattcieslak would you be willing to take a look? The build_log is raising an error, but for some reason that doesn't stop the actual workflow.

tsalo:
Per @mattcieslak's advice, I'll try using `critical` instead of `error`. If that doesn't work, I'll try raising an error (e.g., `ValueError`) after the logged error.

PR #779, opened 2023-01-30
Use constant padding and maximum padlen for temporal filtering

tsalo:
Closes None.

- Replace "odd" padding with "constant", as this will be more similar to the zero padding used by the ABCD consortium.
- Use the maximum allowed padding length (number of volumes - 1).
- Replace filtering functions' MATLAB-generated test files with internal Python-based tests.
- Move censoring description generation to new function, `xcp_d.utils.confounds.describe_censoring`.
- Use a Butterworth filter for the low-pass motion filter option, per Gratton et al. (2020), rather than a FIR filter.

None


tsalo:
Just need to write out a full list of changes.

PR #783, opened 2023-01-31
Use NaNs for low-coverage parcels

### LINKED ISSUE #777: Ensure that zeroed-out time series/correlation coefficients are set to NaNs
## Summary

`wb_command -cifti-correlate` produces ones in the diagonal, even for time series of all zeros. There may be other edge cases where the zeroed-out time series don't end up as NaNs.

tsalo:
Closes #777.

- Create interface to replace time series of zeros (i.e., parcels with bad coverage) in the parcellated CIFTI files with NaNs. It also replaces the correlation values for those parcels (assuming they're in the same order in the data array) with NaNs.
- Use a time series of NaNs for bad parcels in the parcellated NIFTI results as well.

None.

### DIFF
--- a/xcp_d/interfaces/connectivity.py
+++ b/xcp_d/interfaces/connectivity.py
@@ -128,7 +128,7 @@ class NiftiConnect(SimpleInterface):
         time_series = masker.fit_transform(filtered_file)
 
         # Apply the coverage mask
-        time_series[:, coverage_thresholded] = 0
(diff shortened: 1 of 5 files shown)


tsalo:
I just want to write a test or two that explicitly check for NaNs before merging. Stuck on this whole `get_node` thing though.

tsalo:
Closing in favor of #785.

PR #786, opened 2023-02-06
Replace MultiLabel interpolation with GenericLabel

tsalo:
I'm not sure how I missed this. Last year, in October, we had conversations on Slack about using GenericLabel over MultiLabel, but somehow this wasn't translated to a PR in xcp_d.

- Use GenericLabel instead of MultiLabel when applying transforms to atlas files.
- Replace miscellaneous uses of `niworkflows.interfaces.fixes.FixHeaderApplyTransforms`, `nipype.interfaces.ants.ApplyTransforms`, and `xcp_d.interfaces.connectivity.ApplyTransformsx` with a new interface inheriting from `niworkflows.interfaces.fixes.FixHeaderApplyTransforms`. The only change to the interface is that I've overridden the allowed interpolation options to allow GenericLabel.

None.

### DIFF
--- a/xcp_d/interfaces/ants.py
+++ b/xcp_d/interfaces/ants.py
@@ -3,15 +3,20 @@ import logging
 import os
 
 from nipype.interfaces.ants.base import ANTSCommand, ANTSCommandInputSpec
+from nipype.interfaces.ants.resampling import ApplyTransforms as _ApplyTransforms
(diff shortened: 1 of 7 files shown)


mattcieslak (approved):
Great catch, GenericLabel is the preferred ITK method these days

tsalo:
I think I might need to update ANTS in the Docker image.

PR #788, opened 2023-02-06
Mention cosine regressors in aCompCor boilerplate

### LINKED ISSUE #787: Adding to the Methods boilerplate that cosine regressors are added in aCompCor
When the aCompCor denoising strategy is chosen, xcp_d adds cosine regressors to the confounds, but this is not mentioned in the Methods boilerplate in the final report. This is equivalent to applying a high-pass filter to the data.

I think it would be good to add this, because users might not be aware that these extra regressors are added. 

See this relevant thread: https://neurostars.org/t/high-pass-filtering-using-xcp-d/24528/6

tsalo:
Closes #787.

- Document the use of cosine basis regressors in the boilerplate when aCompCor regressors are included.

None

### DIFF
--- a/xcp_d/utils/confounds.py
+++ b/xcp_d/utils/confounds.py
@@ -284,6 +284,8 @@ def describe_regression(params, custom_confounds_file):
             "were selected as nuisance regressors [@behzadi2007component], "
             "along with the six motion parameters and their temporal derivatives "
             "[@benchmarkp;@satterthwaite_2013]. "
+            "As the aCompCor regressors were generated on high-pass filtered data, "
+            "the associated cosine basis regressors were included as well."
         ),
         "acompcor_gsr": (
             "The top 5 aCompCor principal components from the WM and CSF compartments "
@@ -291,6 +293,8 @@ def describe_regression(params, custom_confounds_file):
             "along with the six motion parameters and their temporal derivatives, "
             "mean white matter signal, mean CSF signal, and mean global signal "
             "[@benchmarkp;@satterthwaite_2013]. "
+            "As the aCompCor regressors were generated on high-pass filtered data, "
+            "the associated cosine basis regressors were included as well."
         ),
         "aroma": (
             "AROMA motion-labeled components [@pruim2015ica], mean white matter signal, "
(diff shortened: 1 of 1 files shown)


tsalo:
@leotozzi88 WDYT of the proposed change?

leotozzi88:
I think it would be beneficial to also explain when the data is filtered (at which stage of the denoising pipeline) and what filter is used (i.e. how much is the bandpass?). Also, it might be useful to say explicitly that adding the cosine regressors is equivalent to doing this bandpass filtering.

tsalo:
>I think it would be beneficial to also explain when the data is filtered (at which stage of the denoising pipeline) and what filter is used (i.e. how much is the bandpass?). 

That is mentioned when the bandpass filter is applied. As an example, [here](https://output.circle-artifacts.com/output/job/5841e567-6a78-458e-b733-a73beec8f070/artifacts/0/src/xcp_d/.circleci/out/test_ds001419_nifti/xcp_d/sub-01.html#boilerplate) is the boilerplate for one of the tests. Filtering is mentioned after interpolation.

>Also, it might be useful to say explicitly that adding the cosine regressors is equivalent to doing this bandpass filtering.

AFAIK, the cosine regressors act as a high-pass filter, rather than a bandpass filter, so they're not equivalent. I'd love to put the cutoff value in the boilerplate, but I don't know that it's something that can be extracted from the fMRIPrep derivatives. The only place I've found reference to the exact cutoff is in the fMRIPrep code, where it appears to be hardcoded. I'm not really comfortable hardcoding a value into the boilerplate, when it could be changed in the fMRIPrep source code without me noticing.

leotozzi88:
Yes sorry I meant high-pass not bandpass. But I see your point about not hard coding, that makes sense!

tsalo:
Would something like the following work?

>As the aCompCor regressors were generated on high-pass filtered data, the associated cosine basis regressors were included as well. **This has the effect of high-pass filtering the data as well.**

leotozzi88:
Yes I think that clarifies what is happening thank you so much!

Leonardo Tozzi

From: Taylor Salo ***@***.***>
Date: Tuesday, February 7, 2023 at 8:56 AM
To: PennLINC/xcp_d ***@***.***>
Cc: Leonardo Tozzi ***@***.***>, Mention ***@***.***>
Subject: Re: [PennLINC/xcp_d] Mention cosine regressors in aCompCor boilerplate (PR #788)

Would something like the following work?

As the aCompCor regressors were generated on high-pass filtered data, the associated cosine basis regressors were included as well. This has the effect of high-pass filtering the data as well.

tsalo:
Awesome. I'll add the extra text and merge this PR. Thanks for opening the issue, it was definitely a gap in the boilerplate.

PR #789, opened 2023-02-07
Use nilearn for denoising 2.0

tsalo:

### DIFF
--- a/xcp_d/utils/utils.py
+++ b/xcp_d/utils/utils.py
@@ -461,3 +461,152 @@ def estimate_brain_radius(mask_file, head_radius="auto"):
         brain_radius = head_radius
 
     return brain_radius
+
+
+def denoise_nifti_with_nilearn(
+    bold_file,
+    mask_file,
+    confounds_file,
+    censoring_file,
+    low_pass,
+    high_pass,
+    TR,
+):
+    """Denoise fMRI data with Nilearn.
+
+    Parameters
+    ----------
+    bold_file : str or niimg
+    mask_file : str
+    confounds : pandas.DataFrame
+    low_pass : float
+    high_pass : float
+    TR : float
+    tmask : str
+    """
+    import os
+
+    from nilearn import maskers
+
+    from xcp_d.utils.utils import _denoise_with_nilearn
+
+    out_file = os.path.abspath("desc-denoised_bold.nii.gz")
+
+    # Use a NiftiMasker instead of apply_mask to retain TR in the image header.
+    # Note that this doesn't use any of the masker's denoising capabilities.
+    masker = maskers.NiftiMasker(
+        mask_img=mask_file,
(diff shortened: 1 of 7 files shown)


xcp_d/utils/utils.py
```diff
@@ -461,3 +472,93 @@ def estimate_brain_radius(mask_file, head_radius="auto"):
...
+    confounds_df_censored = pd.DataFrame(data=confounds_censored, columns=confounds_df.columns)
+
+    # Orthogonalize censored nuisance regressors w.r.t. any censored signal regressors
+    signal_columns = [c for c in confounds_df.columns if c.startswith("signal__")]
+    if signal_columns:
+        warnings.warn(
+            "Signal columns detected. "
+            "Orthogonalizing nuisance columns w.r.t. the following signal columns: "
+            f"{', '.join(signal_columns)}"
+        )
```
tsalo, line 516: Switching to orthogonalization instead of non-aggressive denoising.

xcp_d/utils/utils.py
```diff
@@ -461,3 +472,93 @@ def estimate_brain_radius(mask_file, head_radius="auto"):
...
+        betas = np.linalg.lstsq(signal_regressors, noise_regressors, rcond=None)[0]
+        pred_noise_regressors = np.dot(signal_regressors, betas)
+        orth_noise_regressors = noise_regressors - pred_noise_regressors
+        temp_confounds_df.loc[:, :] = orth_noise_regressors
+        confounds_df_censored = temp_confounds_df
+
+    # Then detrend and regress
+    clean_data_censored = signal.clean(
+        signals=raw_data_censored,
+        detrend=True,
```
tsalo, line 530: Does it make sense to detrend the censored data?

xcp_d/utils/utils.py
```diff
@@ -461,3 +472,93 @@ def estimate_brain_radius(mask_file, head_radius="auto"):
...
+    )
+
+    # Now interpolate with cubic spline interpolation
+    clean_data_interp = np.zeros((n_volumes, n_voxels), dtype=clean_data_censored.dtype)
+    clean_data_interp[sample_mask, :] = clean_data_censored
+    clean_data_interp = signal._interpolate_volumes(
+        clean_data_interp,
+        sample_mask=sample_mask,
+        t_r=TR,
+    )
```
tsalo, line 547: Switching from linear interpolation to cubic spline interpolation. Is this a problem?

xcp_d/utils/utils.py
tsalo: Non-aggressive denoising doesn't work in clean without making changes to nilearn. However, orthogonalization might be the better approach anyway.

tsalo:
no point in doing this when the executive summary requires the uncensored denoised data

Issue #793, opened 2023-02-10
xcp_d cannot access miniconda with singularity


StevenGeysen:
## Summary
Using the singularity image, we run into a permission error `PermissionError: [Errno 13] Permission denied: '/usr/local/miniconda/lib/python3.8/site-packages/.wh.pip-20.2.4-py3.8.egg-info'`. It appears that xcp_d creates a miniconda environment that it cannot access later on.

## Additional details

- xcp_d version: 0.3.2 (latest)

### What were you trying to do?
We tried to apply xcp_d to fMRIPrep output.
### What did you expect to happen?
We expected the scripts to run without errors and end up with regressed data.
### What actually happened?
The process breaks and returns `PermissionError: [Errno 13] Permission denied: '/usr/local/miniconda/lib/python3.8/site-packages/.wh.pip-20.2.4-py3.8.egg-info'`

The full error:
```
Matplotlib created a temporary config/cache directory at /tmp/matplotlib-viusok7q because the default path (/home/xcp_d/.config/matplotlib) is not a writable directory; it is highly recommended to set the MPLCONFIGDIR environment variable to a writable directory, in particular to speed up the import of Matplotlib and to better support multiprocessing.
Traceback (most recent call last):
  File "/usr/local/miniconda/bin/xcp_d", line 8, in <module>
    sys.exit(main())
  File "/usr/local/miniconda/lib/python3.8/site-packages/xcp_d/cli/run.py", line 547, in main
    from xcp_d.utils.sentry import sentry_setup
  File "/usr/local/miniconda/lib/python3.8/site-packages/xcp_d/utils/__init__.py", line 5, in <module>
    from xcp_d.utils import (
  File "/usr/local/miniconda/lib/python3.8/site-packages/xcp_d/utils/concatenation.py", line 13, in <module>
    from nilearn.image import concat_imgs
  File "/usr/local/miniconda/lib/python3.8/site-packages/nilearn/__init__.py", line 39, in <module>
    import pkg_resources
  File "/usr/local/miniconda/lib/python3.8/site-packages/pkg_resources/__init__.py", line 3239, in <module>
    def _initialize_master_working_set():
  File "/usr/local/miniconda/lib/python3.8/site-packages/pkg_resources/__init__.py", line 3222, in _call_aside
    f(*args, **kwargs)
  File "/usr/local/miniconda/lib/python3.8/site-packages/pkg_resources/__init__.py", line 3251, in _initialize_master_working_set
[...]

PR #794, opened 2023-02-10
Change permissions of miniconda folders in Dockerfile

tsalo:
Will hopefully solve #793 🤞.

- Based on https://github.com/nipreps/mriqc/issues/647, `chmod` `/usr/local/miniconda`. This is already done in the `xcpd_build` Dockerfile, but @mattcieslak and I think that the permissions could be reset at some point in the build process.

None


tsalo:
I'm going to close this. The problem stemmed from an old version of Singularity.

Issue #813, opened 2023-02-20
Time series and correlation matrices are interpolated


tsalo:
## Summary

Currently, xcp_d's denoised data, time series, and correlation matrices come from the interpolated data. We need to re-censor these data after bandpass filtering.

## Additional details

- xcp_d version: all
- Docker version: 
- Singularity version:

PR #814, opened 2023-02-20
MAINT 0.3.x: Censor filtered data

tsalo:
References #813.

- Censor filtered BOLD data before parcellating and computing correlation matrices.

The outputs file.


tsalo:
Action items:

1. Write out interpolated data if dcan-qc flag is used, but use a different desc field.
2. Figure out what additional files need to be censored. In some cases, we should write out the full file when the dcan-qc flag is used.
3. Make sure concatenation code separately concatenates the censored and interpolated files.
4. Figure out early stopping rule. Currently, this is 50% high motion volumes, but the UMN folks may want something different.

PR #838, opened 2023-03-23
Support preprocessing derivatives with T2w, but no T1w

### LINKED ISSUE #802: Support T2w-only workflow
## Summary
Nibabies will add a t2w-only workflow when T1w images are not available.

1. Make T1w optional, but require a T2w in that case. Use the T2w in all of the cases where the T1w is used now.
2. Search for T2w-to-template and template-to-T2w transforms when the T1w isn't available.

tsalo:
Closes #802.

- Update `collect_data` to grab the T2w-to-standard, standard-to-T2w, and T2w-to-native transforms when there is no T1w image available.
- Replace "t1w" in many variables with "anat", since those variables may refer to T1w or T2w data.


tsalo:
Fixed.

PR #839, opened 2023-03-23
Parcellate ReHo, ALFF, and surface morphometric maps

### LINKED ISSUE #831: Parcellate resting state derivatives and morphometrics (when available)
## Summary

Ted brought up writing out the parcellated ALFF, ReHo, curv, sulc, and thickness files as TSVs in today's meeting.

tsalo:
Closes #831.

- Add parameter `correlate` to NiftiConnect and CiftiConnect. When it is False, the only derivative that will be returned is the timeseries tsv file, which is what we want for ReHo and ALFF. When it is True, the interfaces will return the timeseries, coverage, and correlation tsv files, and CiftiConnect will also return the CIFTI versions of those files as well.
- Split out the atlas-loading and warping/resampling steps from the connectivity workflows, into `init_load_atlases_wf`.
- Create a new workflow, `init_parcellate_surfaces_wf`, to parcellate the surface morphometric files.

None


tsalo:
The part I'm stuck on is parcellating all combinations of a few surfaces and several atlases.

PR #862, opened 2023-04-18
Support high-pass or low-pass only filtering

### LINKED ISSUE #844: Allow `--lower-bpf` or `--upper-bpf` to be None to enable high-pass/low-pass filters
## Summary

Some users want to be able to just apply a high-pass filter, which seems feasible if we just allow the bandpass filter parameters to be None.

+1 from Ted.

smeisler:

Closes #844. Happy to work on this a bit. Was wondering if you think it would be better to have a separate flag for disabling  the filter components, or rather do something like "set `--lower-bpf` to 0 to disable" (or something else)? What I have begun drafting so far uses a "set to 0" approach.

Can ALFF be run on data that has only been low-passed?

### DIFF
--- a/xcp_d/cli/run.py
+++ b/xcp_d/cli/run.py
@@ -668,9 +668,9 @@ def build_workflow(opts, retval):
     if opts.lower_bpf == 0 and opts.upper_bpf == 0:
         opts.bandpass_filter = False
     if (
-        opts.bandpass_filter and
-        (opts.lower_bpf >= opts.upper_bpf) and
-        (opts.lower_bpf > 0 and opts.upper_bpf > 0)
+        opts.bandpass_filter
+        and (opts.lower_bpf >= opts.upper_bpf)
+        and (opts.lower_bpf > 0 and opts.upper_bpf > 0)
     ):
         build_log.error(
             f"'--lower-bpf' ({opts.lower_bpf}) must be lower than "
--- a/xcp_d/cli/run.py
+++ b/xcp_d/cli/run.py
@@ -665,6 +665,8 @@ def build_workflow(opts, retval):
         retval["return_code"] = 1
 
     # Bandpass filter parameters
+    if opts.lower_bpf == 0 and opts.upper_bpf == 0:
+        opts.bandpass_filter = False
     if (
         opts.bandpass_filter and
         (opts.lower_bpf >= opts.upper_bpf) and
--- a/xcp_d/cli/run.py
+++ b/xcp_d/cli/run.py
@@ -291,7 +291,7 @@ def get_parser():
         type=float,
         help=(
             "Lower cut-off frequency (Hz) for the Butterworth bandpass filter to be applied to "
-            "the denoised BOLD data. "
+            "the denoised BOLD data. Set to 0.0 or negative to disable high-pass filtering. "
             "See Satterthwaite et al. (2013)."
         ),
     )
@@ -302,7 +302,7 @@ def get_parser():
         type=float,
         help=(
             "Upper cut-off frequency (Hz) for the Butterworth bandpass filter to be applied to "
-            "the denoised BOLD data. "
+            "the denoised BOLD data. Set to 0.0 or negative to disable low-pass filtering. "
             "See Satterthwaite et al. (2013)."
         ),
     )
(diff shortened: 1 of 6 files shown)


smeisler:
> Regarding ALFF... apparently ALFF properly only refers to data filtered between .01 and .08 Hz, so it looks like any bandpass/lowpass/highpass parameters outside of that range can still produce something useful, but that output should have a different suffix. 

Interesting, good to know. Will keep this PR just to to the filtering functionality then.

> What do you think? As a user, would having a --lowpass alias (and --highpass of course) for the bpf parameters make it clearer that these parameters aren't necessarily for a bandpass filter? Or would additional aliases just muddy the waters?


I think as long as the documentation makes it clear that either high- or low-pass can be disabled with the current naming conventions, it is not worth adding additional aliases or making any breaking changes.

xcp_d/utils/utils.py
```diff
@@ -313,10 +313,22 @@ def butter_bandpass(
...
 
+    if low_pass == 0 and high_pass > 0:
+        btype = "highpass"
+        filt_input = high_pass
+    elif low_pass > 0 and high_pass == 0:
+        btype = "lowpass"
+        filt_input = low_pass
+    elif low_pass > 0 and high_pass > 0:
+        btype = "bandpass"
+        filt_input = [high_pass, low_pass]
```
tsalo, line 324: I think this could be simplified if you put bandpass at the top. E.g., `if low_pass > 0 and high_pass > 0... elif low_pass > 0... elif high_pass > 0`. I don't think we need to worry about negative values (i.e., they should just be treated as 0s).

tsalo (commented):
Can you also test out the highpass and lowpass behaviors in the CLI tests (one test for each new behavior would be great), and make sure that the sidecar files correctly indicate the type of filter and its settings?

xcp_d/utils/utils.py
smeisler: done

tsalo: thx!

smeisler:
> Can you also test out the highpass and lowpass behaviors in the CLI tests (one test for each new behavior would be great), and make sure that the sidecar files correctly indicate the type of filter and its settings?

Will do, but I am having some trouble finding which code is responsible for writing the JSONs out. I have changed the documentation and boilerplate text accordingly.

smeisler:
Great, all the commits should be populating here momentarily.

smeisler:
Cannot seem to get the run.py changes to be in a way that BLK likes, but besides that I think everything is there.

tsalo:
>Cannot seem to get the run.py changes to be in a way that BLK likes, but besides that I think everything is there.

That sometimes happens if my local version of black is out-of-date, since XCP-D's current setup will just use the most recent version.

Regarding the tests, I think I may have misrepresented what I was thinking of. I think adding two new tests is overkill. Instead, can you just modify two of the existing tests to use the new settings? So one should have a bandpass filter (this should already exist), one should have a lowpass, one should have a highpass, and one should disable the filter completely (this should already exist). There are 5 workflow/CLI tests, so the last one can just keep the bandpass filter.

xcp_d/cli/run.py
```diff
@@ -665,7 +665,8 @@ def build_workflow(opts, retval):
         retval["return_code"] = 1
 
     # Bandpass filter parameters
-    if opts.bandpass_filter and (opts.lower_bpf >= opts.upper_bpf):
+    if opts.bandpass_filter and (opts.lower_bpf >= opts.upper_bpf) and \
+            (opts.lower_bpf > 0 and opts.upper_bpf > 0):
```
tsalo, line 669: It could also be that black flags a style issue, but isn't able to solve it. I think black would generally recommend something like the following though:

```suggestion
    if (
        opts.bandpass_filter and
        (opts.lower_bpf >= opts.upper_bpf) and
        (opts.lower_bpf > 0 and opts.upper_bpf > 0)
    ):
```

smeisler:
Updated tests accordingly

tsalo:
Is your PR ready for a final review? If so, can you remove the prefixes from the PR title and update your branch from `main`?

tsalo (approved):
The changes look good to me. Thanks for tackling this!

PR #870, opened 2023-04-25
Add additional argument aliases

smeisler:
I realized `mem-GB` was not recognized while `mem_GB` was. So I made sure dash and underscored aliases were available for named arguments.

Change run.py to have additional aliases.


tsalo (approved):
All of these changes look good to me. Thanks for the PR!

PR #872, opened 2023-04-30
Fix issues with HCP Ingression

kahinimehta:
"sub-" prefix not being added to subject directories in BIDS style derivatives folder


tsalo (approved):
The changes look good to me.

PR #879, opened 2023-05-18
Add "none" option for denoising

### LINKED ISSUE #845: [DCAN] Add "none" option for nuisance regression strategy
## Summary

Users have requested the ability to run XCP-D without denoising (e.g., #775). Adding a nuisance regression strategy that doesn't perform nuisance regression could serve this purpose.

+1 from Ted.

tsalo:
Closes #845.

- Add option to skip denoising (`--nuisance-regressors none`).


xcp_d/tests/test_confounds.py
```diff
@@ -52,7 +52,7 @@ def test_custom_confounds(fmriprep_with_freesurfer_data, tmp_path_factory):
         custom_confounds=custom_confounds_file,
     )
     # We expect n params + 2 (one for each condition in custom confounds)
-    assert combined_confounds.shape == (N_VOLUMES, 26)
+    assert combined_confounds.shape == (N_VOLUMES, 28)
```
mattcieslak, line 55: Why did this change?

xcp_d/tests/test_interfaces_censoring.py
```diff
@@ -25,7 +27,7 @@ def test_generate_confounds(fmriprep_with_freesurfer_data, tmp_path_factory):
     df.to_csv(confounds_tsv, sep="\t", index=False, header=True)
 
     # Run workflow
-    interface = GenerateConfounds(
+    interface = censoring.GenerateConfounds(
```
mattcieslak, line 30: is this for readability?

xcp_d/utils/confounds.py
```diff
@@ -437,6 +443,9 @@ def load_confound_matrix(
         custom_confounds_df = pd.read_table(custom_confounds, sep="\t")
         confounds_df = pd.concat([custom_confounds_df, confounds_df], axis=1)
 
+    confounds_df["linear_trend"] = np.arange(confounds_df.shape[0])
```
mattcieslak, line 446: Does this need to be centered and scaled?

xcp_d/workflows/postprocessing.py
```diff
@@ -290,42 +290,43 @@ def init_prepare_confounds_wf(
...
 
-    plot_design_matrix = pe.Node(
-        niu.Function(
-            input_names=["design_matrix", "temporal_mask"],
-            output_names=["design_matrix_figure"],
-            function=_plot_design_matrix,
-        ),
-        name="plot_design_matrix",
-    )
+    if params != "none":
```
mattcieslak, line 293: Amazing how many changes it takes to make xcpd not denoise. This looks good though

mattcieslak (approved):

xcp_d/tests/test_confounds.py
tsalo: I moved the addition of the linear trend and intercept columns into `load_confound_matrix`. Before they were added on afterwards, so the output of `load_confound_matrix` has two more columns.

xcp_d/utils/confounds.py
tsalo: I center all of the confounds (except the intercept, of course) after censoring. I figured that would be the way to go, but I can change it to using the full confounds if that's better.

xcp_d/tests/test_interfaces_censoring.py
tsalo: Yes. I like to import the module for these module-level test files, so I don't have to import the individual classes and functions that will be tested.

xcp_d/utils/confounds.py
mattcieslak: Your approach is better, I forgot about the possibility of censoring

tsalo:
The executive summary's postprocessing carpet plot looks bad because no centering/detrending is applied, but I don't know if that's something that should be fixed or just accepted.

mattcieslak:
Do you have an example of the bad ES carpetplots? We can add them to one of the DCAN issues

tsalo:
Here's an example from a recent test run:

image

The issue is that, when denoising is applied, the postprocessed data are already mean-centered and detrended, so the carpet plot looks alright. Also, the DCAN team wants a specific range of values for the colorbar.

When denoising isn't applied, the carpet plot doesn't have mean-centered or detrended data. The preprocessing data are centered and detrended right before the carpet plot is made, but not the postprocessed data. We could easily apply the same step to the postprocessed data, but I would need to add a `params` argument to the interface to make it dependent on denoising not being applied.

tsalo:
I opened a separate issue about the figure, so I'm going to merge this now.

PR #882, opened 2023-05-30
Update docs

kahinimehta:
Make changes to docs based on Arielle's comments

### DIFF
--- a/docs/workflows.rst
+++ b/docs/workflows.rst
@@ -119,7 +119,7 @@ Framewise displacement calculation and thresholding
 :class:`~xcp_d.interfaces.censoring.GenerateConfounds`,
 :func:`~xcp_d.utils.modified_data.compute_fd`
 
-Framewise displacement is then calculated according to the formula from Power et al. (CITE).
+Framewise displacement is then calculated according to the formula from Power et al. :footcite:t:`power_fd_dvars`.
 Two parameters that impact FD calculation and thresholding are
 (1) the head radius used to convert rotation degrees to millimeters and
 (2) the framewise displacement threshold.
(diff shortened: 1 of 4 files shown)


docs/workflows.rst
```diff
@@ -119,7 +119,7 @@ Framewise displacement calculation and thresholding
 :class:`~xcp_d.interfaces.censoring.GenerateConfounds`,
 :func:`~xcp_d.utils.modified_data.compute_fd`
 
-Framewise displacement is then calculated according to the formula from Power et al. (CITE).
+Framewise displacement is then calculated according to the formula from Power et al. :footcite:t:`power_fd_dvars`.
```
tsalo, line 122:
```suggestion
Framewise displacement is then calculated according to the formula from :footcite:t:`power_fd_dvars`.
```

tsalo: The `:t:` (instead of `:p:`) means that it's an in-text citation, so we don't need the authors beforehand.

xcp_d/interfaces/nilearn.py
```diff
@@ -66,7 +66,8 @@ class _SmoothInputSpec(BaseInterfaceInputSpec):
             minlen=3,
             maxlen=3,
         ),
-        desc="Smoothing strength, as a full-width at half maximum, in millimeters.",
+        desc="Full width at half maximum. \
+        Smoothing strength, as a full-width at half maximum, in millimeters.",
```
tsalo, line 70:
```suggestion
        desc=(
            "Full width at half maximum. "
            "Smoothing strength, as a full-width at half maximum, in millimeters."
        ),
```

tsalo (changes requested):
I have a few minor suggestions, but otherwise it looks really good. Thank you.

tsalo (approved):
LGTM thx!

PR #885, opened 2023-06-02
Add `--exact-time` parameter

### LINKED ISSUE #883: [DCAN] Add option to retain certain amount of time
## Summary

The DCAN team wants an option to retain a specific amount of time in the outputs (e.g., time series and correlation matrices), as done in [Feczko et al. (2021)](https://www.biorxiv.org/content/10.1101/2021.07.09.451638v1).

The parameter can be called `--exact-time`, and it will be mutually exclusive with `--min-time`.

## Open questions

- Do we want to output both the exact-time time series/matrices and the regular censored ones, or just the exact-time ones if `--exact-time` is used?
- If the DCAN team wants to use `--exact-time`, do we need to write out the interpolated time series?
- Is the variance associated with subsampling important? Should we calculate the correlation matrices for a number of permutations?

## Additional details

1. If the amount of good quality data is less than `--exact-time`, then the run won't be postprocessed at all, as with `--min-time`.
2. The random set of censored volumes will be stored in the temporal mask file, in a new column called `random_censor`.
3. The censored volumes will be plotted in the design matrix and censoring plots. In the censoring plot, it'll have a different color than the high-motion outliers.

tsalo:
Closes #883.

The only thing I'm waiting on is feedback from the DCAN team on my changes to the documentation.

- Add an `--exact-time` parameter.
- Add a `--random-seed` parameter that controls the random seed used when selecting volumes for the `exact-time` correlations.
- Include the randomly censored volumes in the censoring plot.


tsalo:
I'm going to merge as-is, but would love feedback on the `exact-time` documentation regardless.

PR #901, opened 2023-06-19
Fisher-transform connectivity matrices

### LINKED ISSUE #865: Save z-transformed correlation matrices instead of correlation coefficients
## Summary

This is a request from a user (see [this NeuroStars post](https://neurostars.org/t/question-about-output-of-xcp-d/24244/7?u=tsalo)). This is a straightforward transform, so I don't see a reason to write out both r and z values, but z values are probably more useful to users, so it seems like a worthwhile change.

smeisler:
Closes #865. Applies `np.arctanh` to correlation matrices before saving them out. `pearsoncorrelation` is changed to `fishercorrelation` in output filenames.

`outputs.rst
`


tsalo:
@smeisler thanks for opening this! I think the tests are currently failing just because they expect different outputs. Can you update the expected outputs in `xcp_d/tests/data/test_*_outputs.txt`?

Also, one reason we've waited to implement this is that @ericfeczko raised concerns about users assuming the z-transformed coefficients are normally distributed, when there is no guarantee they will be. I think we may need to add a "warning" or "important" admonition box to the section of the documentation that describes the connectivity outputs. Fez linked to this CV post: https://stats.stackexchange.com/a/172422/53066.

smeisler:
@tsalo sounds good, updated the test outputs and rebased from main.

tsalo:
Thanks @smeisler. I'll merge this once CI passes.

tsalo:
I just introduced a bunch of conflicts with a separate PR. I'm sorry about that. You can address the conflicts by accepting all of the incoming changes from `main` and just replacing `pearsoncorrelation` with `fishercorrelation`.

I'm not sure what to do with the diagonals of the correlation matrices though. I guess we could use `n/a`... do you know if the connectivity derivatives BEP has a recommendation for this?

smeisler:
@tsalo I think I took care of the merge, and I asked one of the BEP leaders about the diagonals, will let you know.

tsalo:
Thanks @smeisler!

smeisler:
For the connectivity BEP, it looks like there will be a [JSON field for specifying whether the diagonals contain informative values or not](https://docs.google.com/document/d/1ugBdUF6dhElXdj3u9vw0iWjE6f_Bibsro3ah7sRV0GA/edit#heading=h.s3amy473k9hq) (for example, XCP_D connectivity matrices do not have informative diagonals). In terms of what values go in there, there is no recommendation yet. I would personally go with 0s, as I imagine `nan`s could lead to errors if someone is doing mathematical operations on the matrices, depending on the software.

smeisler:
Added the 0 diagonals and updated the tests, not sure where in the code I would introduce a new JSON output though (`/workflows/outputs.py`?)

tsalo:
I think you could put it toward the top of `init_postproc_derivatives_wf`. I'm not sure if `DerivativesDataSink` can write out _just_ a JSON file, but you could write it out with `json.dump` and a hardcoded filename, as we do in [`write_dataset_description()`](https://github.com/PennLINC/xcp_d/blob/9e3d22c35a1eb1cc1ecfdd29b3bb9cc5f46403b0/xcp_d/utils/bids.py#L693).

smeisler:
A problem with making one overarching `measure-fishercorrelation_conmat.json` in the derivatives root is that `Atlas` is a required field, so we would probably need one JSON per atlas connectivity matrix if that is the final form of the BEP.

smeisler:
The easiest way for me to do this would probably be something like this:

 ```
       if correlate:
            correlations_df.to_csv(
                self._results["correlations"],
                sep="\t",
                na_rep="n/a",
                index_label="Node",
            )
            coverage_df.to_csv(self._results["coverage"], sep="\t", index_label="Node")

            ### NEW CODE ADDITION
            connectivity_json = self._results["correlations"].replace(".tsv",".json")
            connectivity_metadata = {} # Add the BIDS BEP json fields here, changing with different atlases
            write_json(connectivity_metadata, connectivity_json)
```

I understand it may not be the cleanest way though. Would there be a different way that would be more in line with how XCP_D currently outputs accompanying JSONs?

tsalo:
That's unfortunate. How about adding a set of `atlas-<atlasname>_conmat.json`s as well? That would still result in fewer sidecars than having one for every conmat file.

tsalo:
Your proposal would unfortunately only write out a JSON to the working directory for that workflow node.
If we end up writing out a sidecar for every conmat, then the way to go is probably to do what we do with the temporal mask metadata. Namely, add a dictionary output for that node named something like "correlations_metadata" and pass that along to the `postprocessing_derivatives_wf`, at which point you can feed the dictionary directly into the appropriate `DerivativesDataSink` via the `meta_dict` parameter.

smeisler:
I suppose that should work, and result in fewer JSONs if one has multiple runs and/or sessions. I know the BIDS validator inheritance checking is not robust right now but that shouldn't be a concern for this right now.

tsalo:
@smeisler sorry that I didn't follow up and that `main` has drifted so much. I'm just going to close this and open a new one if we ever decide to go forward with it.

PR #905, opened 2023-06-21
Remove workaround for nonbinary Nibabies brain masks

### LINKED ISSUE #904: Remove nibabies mask workaround
## Summary

Nibabies has made a few releases since they fixed the mask datatype bug, so we can remove the workaround we have in place for that.

tsalo:
Closes #904.


tsalo:
I need to update the Nibabies test dataset.

PR #908, opened 2023-06-21
Update Landing Page Figure

kahinimehta:
Figure used in paper


tsalo (commented):
The figure looks good to me, but why not completely replace the old figure? Adding a separate file will increase the size of the repo.

tsalo (approved):
lgtm

Issue #895, opened 2023-06-13
Reach 90% test coverage


tsalo:
## Summary

Stems from #892, as part of the NMIND checklist.

Related to #544 and #506.

tsalo:
The main culprits are `xcp_d.cli` and `xcp_d.utils`.

Within `xcp_d.utils`, the conversion modules (dcan2fmriprep, hcp2fmriprep, and ingestion) all have low coverage. execsummary, filemanip, sentry, and utils are also pretty low.

filemanip, utils, and execsummary should all be reasonably straightforward to test. I am more concerned about testing cli and the conversion modules.

tsalo:
I don't think I'll be able to reach 90% by the manuscript, so I think we'll have to suffice with 80%.

PR #913, opened 2023-06-22
Expand tests

tsalo:
References #895.

- Remove unused functions/classes:
    - _warn_redirect
    - _add_subject_prefix
    - scale_to_min_max
    - _DeprecatedStoreAction040
- Add tests for functions in `xcp_d.utils.bids`.
- Fix minor bug with bbregister plot identification. It looks like, at least in the test datasets, the private OSX files (e.g., `/path/._filename`) were being selected instead of the real files.
    - I still want to replace this step with a better BIDSLayout call, and bundled into an interface, but this PR's changes at least improve the code somewhat.
- Fix bugs in `init_plot_custom_slices_wf` and add a test for it.


tsalo:
I don't know why, but for some reason calling `xcp_d.cli.run.main` directly doesn't work. The actual workflow run seems to go fine, but the test still fails. 🤷

PR #928, opened 2023-07-18
Add AtlasPack atlases

tsalo:


tsalo:
There are currently two problems:

1. Some of the nifti atlases have non-integer values (e.g., 2.1675).
2. The cifti atlases' internal node names differ from the names in the tsv files.

tsalo:
The atlases look solid now. I just need to update the pytests.

tsalo:
The tests are passing now.

PR #956, opened 2023-10-03
Use Nipreps-style configuration object throughout package

### LINKED ISSUE #816: Adopt configuration approach from nipreps tools
## Summary
In the current nipreps tools, parameters and loggers are all centralized in a config object that can be referenced across the library. This seems like something worth doing in xcp_d.

## Additional details

This would bring our code more in line with the nipreps ecosystem. It should also reduce maintenance burden, since many parameters would be centralized.

tsalo:
Closes #816.

- Replace workflow parameters with a config object, when feasible.


tsalo:
Now I'm getting all sorts of errors. Not sure what's going on.

PR #966, opened 2023-10-10
Track `Sources` for outputs with BIDS URIs

tsalo:
Closes #568.

## Breaking changes

- Don't write out the temporal mask if `fd_thresh` is <= 0 (i.e., censoring is disabled).
- Don't write out design file if `params` is None (i.e., denoising is disabled).

- Plug BIDS URIs into appropriate JSON files with the `Sources` field.
- Also add in a number of metadata fields required or recommended by BEP 012 (https://github.com/bids-standard/bids-specification/pull/519).
- Add `DatasetLinks` field to dataset description file. The links are to the preprocessed data, the XCP-D derivatives, and (optionally) the custom confounds.
- Create a series of small functions (listed below) to generate BIDS URIs for different file locations:
    - `_make_uri`: General path-resolving function.
    - `_make_xcpd_uri`: Resolves path for XCP-D derivative files.
    - `_make_xcpd_uri_lol`: Resolves path for XCP-D derivative files organized in lists of lists..
    - `_make_preproc_uri`: Resolves path for preprocessing pipeline derivative files.
    - `_make_custom_uri`: Resolves path for custom confounds files.
- Create some functions that help with managing the `Sources` field in workflows.
    - `_listify`: From NiMARE. Convert things to lists.
    - `_make_dictionary`: Feed kwargs into a dictionary. I use this to create metadata dictionaries containing `Sources` fields to pass into `DerivativesDataSinks`. The issue is that `MapNodes` need to reference fields in the `.inputs` attribute of the  underlying interface, but `DerivativesDataSinks` don't include `Sources` as a field in `.inputs`. However, they _do_ include `meta_dict`, so I need to pass in a `meta_dict` containing the `Sources` field instead of passing in `Sources` directly.
    - `_transpose_lol`: Transposes a list of lists. Used in `_make_xcpd_uri_lol`.


tsalo:
The concatenated atlas-wise sources aren't right. Instead of selecting the appropriate atlas's element in the run-wise lists of sources, it's just keeping the full run-wise lists.

tsalo:
I'm struggling to figure out how to pass `Sources` into a `DerivativesDataSink` as an `iterfield`. The field needs to be available in the node's `inputs` attribute, but only entities from the config file show up there.

tsalo:
@mattcieslak would you mind reviewing this with an eye toward streamlining the code? I'm sure I'll be able to improve it over time, but if there's anything obvious that I'm missing it would be great to catch it before I merge the PR.

PR #979, opened 2023-10-24
Don't convert CIFTIs to int16

### LINKED ISSUE #977: Correlation matrices in CIFTI test results are empty
## Additional details

- xcp_d version:
- Docker version:
- Singularity version:

tsalo:
Closes #977. The problem is that the CIFTI atlas files' contents are being converted to floats.

- Stop using DerivativesDataSink to write out atlas files. There is a bug in the base DerivativesDataSink where dlabel.nii CIFTIs with non-int16 data fail to convert to int16. Before, I was using nibabel to convert the datatype before feeding the file into the DerivativesDataSink. However, there appears to be a new bug in nibabel where that doesn't work, and changing the datatype in the header actually changes the data array's data type from normal ints to weird floats (e.g., 1 --> 1.46867437), which causes problems when we want to actually use those files.
    - Now, I have a function to compile the right filename and copy atlas file to the output folder.
- Also, replace the dseg.tsv file for the Gordon atlas, so the labels are more informative (i.e., match the names from `Gordon333_FreesurferSubcortical.32k_fs_LR.dlabel.nii` from https://sites.wustl.edu/petersenschlaggarlab/parcels-19cwpgu/). This also adds the community labels.
    - I couldn't find communities for Tian or HCP. The 4S atlases have community info for the Schaefer nodes.
- Replace the dseg.tsv file for the Glass atlas to add communities as well. Thanks to Jakob Seidlitz for the communities.


tsalo:
I need to also add community affiliations to the Glasser dseg file, but there's an issue with the label files, so I'll circle back tomorrow.

tsalo:
I just need to write a test for the new copy_atlas function.

PR #982, opened 2023-10-27
Try to reduce memory requirements of functional connectivity nodes

tsalo:
Related to https://github.com/PennLINC/informatics_misc/issues/93.

- Request 15 * the image size for functional connectivity node memory (up from 5 * image size).
- Use `gc.collect()` to remove objects as they stop being used in the functional connectivity nodes.


tsalo:
I don't think this costs us anything, so I'm going to merge it.

PR #1022, opened 2023-12-13
Add options for UK Biobank data

tsalo:
Closes none.

To do:

- [x] Create test dataset. I will need to simulate it, since I can't share real UK Biobank data.
    - I've reduced a PNC subject's data to the first 50 volumes and am now running fMRIPrep 23.2.0a3 on the modified data.
- [x] Add UKB test.
- [ ] ~~Limit height of design matrix, because it ends up being massive when there are only a few columns (e.g., when using the `gsr_only` option).~~ Going to wait for nireports (see #1031).

- Add `--nuisance-regressors` option for GSR only (`gsr_only`), since UK Biobank fMRI data are already denoised with FSL FIX.
- Add `--input-type` option for UK Biobank data (`ukb`).
- Add ingression functions for UK Biobank data.


tsalo:
The normalization in the test data is bad. Will need to fix before merging.

tsalo:
check_affines is failing because the "preprocessed data" folder is not BIDS compliant. I need to pass in the path to the converted BIDS dataset.

PR #1030, opened 2024-01-09
DNM: Add fALFF, PerAF, and tSNR outputs

tsalo:
Add resting-state metric outputs.

- Retain fitted intercept parameter estimates from denoising step as "mean BOLD" image, for later PerAF and tSNR calculation.
- Pass the `uncensored_denoised_bold` data to the ALFF calculation node instead of `interpolated_filtered_bold`, so ALFF and fALFF aren't calculated from bandpass filtered data.
- Add tSNR, PerAF, and fALFF outputs.


xcp_d/utils/restingstate.py
```diff
@@ -133,21 +141,25 @@ def compute_alff(data_matrix, low_pass, high_pass, TR, sample_mask=None):
...
+            voxel_data_for_peraf = voxel_data_censored.copy()
+
+            # Normalize data matrix over time. This will ensure that the standard periodogram and
+            # Lomb-Scargle periodogram will have the same scale.
+            voxel_data_censored -= np.mean(voxel_data_censored)
+            voxel_data_censored /= np.std(voxel_data_censored)
+
             time_arr = np.arange(0, n_volumes * TR, TR)
             assert sample_mask.size == time_arr.size, f"{sample_mask.size} != {time_arr.size}"
             time_arr = time_arr[sample_mask]
```
mattcieslak, line 165: 0.5 * fs is the nyquist frequency?

xcp_d/utils/restingstate.py
```diff
@@ -178,10 +197,22 @@ def compute_alff(data_matrix, low_pass, high_pass, TR, sample_mask=None):
...
+            np.mean(np.abs((voxel_data_for_peraf - mean_matrix[i_voxel]) / mean_matrix[i_voxel]))
+            * 100
+        )
+
+        tsnr[i_voxel] = mean_matrix[i_voxel] / np.std(voxel_data_for_peraf)
 
     assert alff.size == n_voxels, f"{alff.shape} != {n_voxels}"
 
-    # Add second dimension to array
+    # Add second dimension to arrays
```
mattcieslak, line 212: I like to use `.reshape(-1, 1)` for padding, but I don't think there's an official best way to do it

xcp_d/utils/restingstate.py
```diff
@@ -178,10 +197,22 @@ def compute_alff(data_matrix, low_pass, high_pass, TR, sample_mask=None):
...
         alff[i_voxel] = len(ff_alff) * np.mean(power_spectrum_sqrt[ff_alff[0] : ff_alff[1]])
+        falff[i_voxel] = np.sum(power_spectrum_sqrt[ff_alff[0] : ff_alff[1]]) / np.sum(
+            power_spectrum_sqrt
+        )
+        peraf[i_voxel] = (
+            np.mean(np.abs((voxel_data_for_peraf - mean_matrix[i_voxel]) / mean_matrix[i_voxel]))
+            * 100
+        )
+
+        tsnr[i_voxel] = mean_matrix[i_voxel] / np.std(voxel_data_for_peraf)
```
mattcieslak, line 208: Does this ever produce nans or infs?

mattcieslak (commented):
The only thing I'd be sure to add is a check for nan/inf values in percent alff

xcp_d/utils/restingstate.py
tsalo: I have a check above for a zero-variance time series that will set the voxel's value to 0 instead of running the equation, so it shouldn't produce any NaNs or Infs.

mattcieslak: sorry I missed it! In that case I approve

mattcieslak (approved):

xcp_d/interfaces/nilearn.py
```diff
@@ -336,6 +341,14 @@ def _run_interface(self, runtime):
             TR=self.inputs.TR,
         )
 
+        self._results["mean_bold"] = os.path.join(runtime.cwd, "mean.dtseries.nii")
+        write_ndata(
+            mean_bold,
+            template=self.inputs.preprocessed_bold,
+            filename=self._results["mean_bold"],
```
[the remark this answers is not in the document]
tsalo: It's created automatically by the class, I believe. I define `self._results["mean_bold"]` in the line before this I think (344).

xcp_d/utils/restingstate.py
tsalo: Yes it should be.

xcp_d/utils/restingstate.py
tsalo: I actually think the Numpy-preferred method might be [`numpy.expand_dims`](https://numpy.org/doc/stable/reference/generated/numpy.expand_dims.html), but I don't feel strongly about it.

Issue #1032, opened 2024-01-12
Scaling data dramatically affects ALFF


tsalo:
## Summary

ALFF is dependent on the scale of the data, so the scaling step I added in #1020 dramatically changes the ALFF outputs. Unfortunately, it's not just a change in the ALFF scale, since some testing by @joellebagautdinova shows that the ALFF maps produced from scaled data don't show any of the spatial patterns that ALFF maps produced from unscaled data do.

The spatial patterns in the unscaled ALFF maps are very similar to the mean BOLD maps from the preprocessed data (and probably the intercept image from the denoising step), so ALFF itself may be problematic, but it's also clear that the scaled ALFF looks bad.

## Additional details

- xcp_d version: 0.6.1 (unreleased)
- Docker version:
- Singularity version:

tsalo:
The scaling _difference_ is handled by #1033, but my concerns about ALFF remain unresolved.

tsalo:
Should we do what other ALFF implementations have done and scale by the mean ALFF value in the brain?

tsalo:
>What is the best way to handle the ALFF output?

The short answer is I don't know. I've had this issue open for over year (yikes!) because I don't have a good answer. The best solution I've come across is to divide the vertex-wise ALFF values by the mean ALFF in the brain, but I haven't personally tried that out. Is there a reason you tried it out yourself and decided not to use that approach?

>I tried to rescale it using the standard deviation of the BOLD fluctuation for each vertex (fsLR_den-91k_desc-denoised_bold.dtseries.nii) - by dividing the ALFF measure of each vertex (fsLR_den-91k_stat-alff_boldmap.dscalar.nii) by the BOLD SD. Looking at the xcp-d code, I believe this is the scaling that was applied to the ALFF data and I wanted to "revert" it.

As I'm sure you saw, the spatial patterns you'd expect to see in ALFF basically go away when you do that.

>However, this also raised the question for me as to why scaling is used at all, especially as it was not implemented in the 0.4 version used in the paper I referenced.

[...]

PR #1033, opened 2024-01-16
Rescale ALFF based on original BOLD standard deviation

tsalo:
Related to #1032.

- Calculate the SD of the denoised BOLD data, then multiply the ALFF computed on the scaled BOLD data by that SD. This brings ALFF back to its original scale, and the censored version of ALFF will has the same scale as the uncensored version.

### DIFF
--- a/docs/workflows.rst
+++ b/docs/workflows.rst
@@ -474,15 +474,22 @@ spontaneous neural activity in resting-state BOLD data.
 It is calculated by the following:
 
 1. The ``filtered, interpolated, denoised BOLD`` is passed along to the ALFF workflow.
-2. Voxel-wise BOLD time series are normalized (mean-centered and scaled to unit standard deviation)
+2. If censoring+interpolation was performed, then the interpolated time series is censored at this
+   point.
+3. Voxel-wise BOLD time series are normalized (mean-centered and scaled to unit standard deviation)
    over time.
-3. The power spectrum and associated frequencies are estimated from the BOLD data.
+4. The power spectrum and associated frequencies are estimated from the BOLD data.
+
    -  If censoring+interpolation was not performed, then this uses :func:`scipy.signal.periodogram`.
    -  If censoring+interpolation was performed, then this uses :func:`scipy.signal.lombscargle`.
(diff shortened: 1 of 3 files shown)


docs/workflows.rst
```diff
@@ -474,15 +474,22 @@ spontaneous neural activity in resting-state BOLD data.
...
-5. The power spectrum values corresponding to the frequency range retained by the
+
+5. The square root of the power spectrum is calculated.
+6. The power spectrum values corresponding to the frequency range retained by the
    temporal filtering step are extracted from the full power spectrum.
-6. The mean of the within-band power spectrum is calculated and multiplied by 2.
+7. The mean of the within-band power spectrum is calculated and multiplied by 2.
+8. The ALFF value is multiplied by the standard deviation of the voxel-wise
+   ``filtered, interpolated, denoised BOLD`` time series.
+   This brings ALFF back to its original scale, as if the time series was not normalized.
```
mattcieslak, line 492: Do you think it would be worth mentioning that this is particularly important for the censored data to match uncensored?

xcp_d/utils/restingstate.py
```diff
@@ -178,6 +184,8 @@ def compute_alff(data_matrix, low_pass, high_pass, TR, sample_mask=None):
         # from the value closest to the low pass cutoff, to the value closest
         # to the high pass pass cutoff
         alff[i_voxel] = len(ff_alff) * np.mean(power_spectrum_sqrt[ff_alff[0] : ff_alff[1]])
+        # Rescale ALFF based on original BOLD scale
+        alff[i_voxel] *= sd_scale
```
mattcieslak, line 188: ohh so both periodogram and lombscargle get the scaling and rescaling?

mattcieslak (approved):

xcp_d/utils/restingstate.py
tsalo: Yes, though it's probably unnecessary for the periodogram version.

docs/workflows.rst
tsalo: Happy to add some text to that effect.

PR #1040, opened 2024-01-22
Adopt Nipreps-style Config object

### LINKED ISSUE #734: Add `--config-file` option
## Summary

There are a lot of command line options, so it might be good to support a configuration file option, where users can write out the options in a JSON file instead of putting them in the command line call.

## Additional details
fMRIPrep allows a config file, so it's a common nipreps argument.

### LINKED ISSUE #816: Adopt configuration approach from nipreps tools

tsalo:
Closes #816, closes #734.

- Consolidate shared parameters in a Config object.
- Consolidate `skip_<integration-test-name>` shortcuts into `skip_integration`.
- Add several BIDS App-related parameters:
    - `--write-graph`
    - `--boilerplate-only`
    - `--debug`
    - `--md-only-boilerplate`
    - `--reports-only`
    - `--low-mem`: Not sure if this works.
    - `--config-file`


tsalo:
I think at this point I just need to update the units tests and add a mock config for the workflow figures in docstrings and the documentation.

tsalo:
Not sure why, but `tests/data/config.toml` isn't being found by RTD. ASLPrep has the same folder structure, but RTD is able to find the test config file for that one, and I can't see any differences in the infrastructure files.

tsalo:
I've decided to just open a separate issue about the workflow documentation.

PR #1054, opened 2024-02-15
Replace locally-stored spheres with TemplateFlow

tsalo:
I've opened https://github.com/templateflow/templateflow/issues/173 requesting the equivalent of `fs_<hemi>-to-fs_LR_fsaverage.<hemi>_LR.spherical_std.164k_fs_<hemi>.surf.gii` be uploaded to TemplateFlow.

to do:

- [x] Bump templateflow Python client version to 24.0.0, since that's when the new templates were added.


xcp_d/interfaces/bids.py
```diff
@@ -129,18 +131,19 @@ def _run_interface(self, runtime):
...
 
-            # TODO: Collect from templateflow once it's uploaded.
-            # FreeSurfer: fs_?/fs_?-to-fs_LR_fsaverage.?_LR.spherical_std.164k_fs_?.surf.gii
-            self._results["sphere_to_sphere"] = pkgrf(
-                "xcp_d",
-                (
-                    f"data/standard_mesh_atlases/fs_{hemisphere}/"
-                    f"fs_{hemisphere}-to-fs_LR_fsaverage.{hemisphere}_LR.spherical_std."
-                    f"164k_fs_{hemisphere}.surf.gii"
-                ),
```
tsalo, line 140: @madisoth is this `tpl-fsLR_space-fsaverage_hemi-<L|R>_den-164k_sphere.surf.gii` (which is on TemplateFlow)? Or would it be `tpl-fsaverage_space-fsLR_hemi-<L|R>_den-164k_sphere.surf.gii` (which I think we would need to add to TemplateFlow)?

xcp_d/interfaces/bids.py
```diff
@@ -165,28 +166,40 @@ def _run_interface(self, runtime):
...
+            self._results["source_sphere"] = str(
+                get_template(
+                    template="fsaverage",
+                    space=None,
+                    hemi=hemisphere,
+                    density="41k",
+                    desc="reg",
+                    suffix="sphere",
+                )
             )
```
tsalo, line 179: There is no `tpl-fsaverage_hemi-<L|R>_den-41k_desc-reg_sphere.surf.gii` on TemplateFlow. Do we need to add this, or is it perhaps just the same as `tpl-fsaverage_hemi-<L|R>_den-41k_sphere.surf.gii`?

xcp_d/interfaces/bids.py
```diff
@@ -165,28 +166,40 @@ def _run_interface(self, runtime):
...
+            self._results["sphere_to_sphere"] = str(
+                get_template(
+                    template="dhcpSym",
+                    space="fsaverage",
+                    hemi=hemisphere,
+                    density="41k",
+                    desc="reg",
+                    suffix="sphere",
+                )
             )
```
tsalo, line 191: We need cohorts here. Also, I believe we decided not to add the warp spheres when adding the templates to TemplateFlow, but it would be good to get them on there soon.

xcp_d/interfaces/bids.py
```diff
@@ -165,28 +166,40 @@ def _run_interface(self, runtime):
...
+            self._results["target_sphere"] = str(
+                get_template(
+                    template="dhcpSym",
+                    space="fsLR",
+                    hemi=hemisphere,
+                    density="32k",
+                    desc="week42",
+                    suffix="sphere",
+                )
             )
```
tsalo, line 203: Replace desc with cohort.

xcp_d/interfaces/bids.py
tsalo: Also, should this really be fsaverage and not dhcpSym?

xcp_d/interfaces/bids.py
tsalo: Currently not found.

PR #1056, opened 2024-02-20
Make description of `smoothing` parameter clearer when set to `0`

psychelzh:
Formerly, for `--smoothing`:

> This may be set to 0.

Such piece of description is a little confusing although one might think smoothing is probably disabled in that case. It could be clearer and more consistent with the description of other parameters, such as `--min-time`, by clearly stating:

> Set to 0 to disable smoothing.

Make the description clearer.

None.


tsalo (approved):
Thanks for opening this! This change looks reasonable to me.

PR #1087, opened 2024-03-12
Adopt Lindquist-compliant denoising method

### LINKED ISSUE #1085: Adopt Lindquist-compatible denoising
## Summary

We currently use what Lindquist 2019 calls MCHP, which they found to be problematic. The Hallquist 2013 paper referred to this as RegBP, and they thought it was fine, but the Lindquist paper is much more convincing.

Since we'll be bandpass filtering before nuisance regression, we won't get out unfiltered, uncensored data (which is used in the executive summary). The DCAN team said they're fine with plotting the interpolated data in the executive summary instead.

Thanks to the manuscript's reviewers for detecting this!

## Additional details

- xcp_d version:
- Docker version:
- Singularity version:

tsalo:
Closes #1085. Replace DCAN's denoising procedure, which is essentially equivalent to RegBP from [Hallquist 2013](https://doi.org/10.1016/j.neuroimage.2013.05.116) or MCHP from [Lindquist 2019](https://doi.org/10.1002/hbm.24528), with an approach that is based on Nilearn's denoising method.

I had to modify the basic structure from the Nilearn code because the DCAN team wants interpolated denoised data, which Nilearn does not provide.

The updated denoising function does the following:

1.  Interpolate high-motion volumes in the BOLD data and confounds.
2.  Detrend interpolated BOLD and confounds.
    -   Only done if denoising is requested.
    -   This also mean-centers the data.
3.  Bandpass filter the interpolated data and confounds.
    -   This is a key element for following Lindquist's recommendations.
5.  Censor the data and confounds.
6.  Estimate betas using only the low-motion volumes.
    -   This is equivalent to denoising based on the low-motion volumes.
7.  Apply the betas to denoise the interpolated BOLD data. This is re-censored in a later step.
    -   The values in the low-motion volumes are the same as if we denoised just the low-motion volumes.


tsalo:
Assuming everything else looks good, I'll need to drop mentions of the mean BOLD signal. ALFF only depends on the SD of the BOLD signal, not the mean.

xcp_d/utils/utils.py
```diff
@@ -333,15 +330,27 @@ def denoise_with_nilearn(
 ):
     """Denoise an array with Nilearn.
 
-    This step does the following:
+    This function is a modified version of Nilearn's signal.clean function, with the following
```
kahinimehta, line 333: Ahh this is super helpful

xcp_d/utils/utils.py
```diff
@@ -380,108 +392,106 @@ def denoise_with_nilearn(
...
+    interpolated_arr = signal._interpolate_volumes(
+        arr,
+        sample_mask=sample_mask,
+        t_r=TR,
+        extrapolate=True,
+    )
+    # Replace any high-motion volumes at the beginning or end of the run with the closest
+    # low-motion volume's data.
+    # Use https://stackoverflow.com/a/48106843/2589328 to group consecutive blocks of outliers.
+    gaps = [[s, e] for s, e in zip(outlier_idx, outlier_idx[1:]) if s + 1 < e]
```
kahinimehta, line 472: Sorry, what are s and e again?

kahinimehta (approved):

xcp_d/utils/utils.py
tsalo: I believe they stand for `start` and `end`. I will update the variable names to make it clearer.

docs/outputs.rst
```diff
@@ -318,10 +318,13 @@ high-motion outliers.
...
 ``design.tsv`` is a tab-delimited file with one column for each nuisance regressor,
-including an intercept column, a linear trend column, and one-hot encoded regressors indicating
-each of the high-motion outlier volumes.
+including one-hot encoded regressors indicating each of the high-motion outlier volumes.
 This file includes the high-motion volumes that are removed in most other derivatives.
 
+.. important::
+   Please note that the outlier columns are somewhat misleading,
+   as volumes are removed by censoring, rather than regression.
+
```
mattcieslak, line 327: Is this also true if despiking is turned on?

docs/workflows.rst
```diff
@@ -359,27 +341,75 @@ Denoising
...
+
+Specifically, temporal filtering and confound regression are performed in separate stages,
+but the confounds are orthogonalized with respect to the temporal filter prior to confound
+regression,
+so that no variance from the temporal filter is reintroduced by the confound regression.
+
+XCP-D modifies Nilearn's approach in the following ways:
+
+1. XCP-D uses :func:`numpy.linalg.lstsq` to estimate betas instead of QR decomposition,
+   in order to denoise the interpolated data as well.
```
mattcieslak, line 356: Do we need to justify this choice?

docs/workflows.rst
```diff
@@ -405,57 +435,27 @@ mean-centered.
...
+as users **should not** use interpolated data directly.
 
 
 Re-censoring
 ------------
 :class:`~xcp_d.interfaces.censoring.Censor`
 
 After bandpass filtering, high motion volumes are removed from the
-``filtered, interpolated, denoised BOLD`` once again, to produce ``filtered, denoised BOLD``.
+``denoised, interpolated BOLD`` once again, to produce ``denoised BOLD``.
```
mattcieslak, line 458: Maybe not for this PR, but I wonder which of these parts deserve special highlighting. maybe in a call-out box

xcp_d/interfaces/concatenation.py
```diff
@@ -279,12 +266,7 @@ class _ConcatenateInputsInputSpec(BaseInterfaceInputSpec):
         mandatory=True,
         desc="TSV files with high-motion outliers indexed.",
     )
-    uncensored_denoised_bold = traits.List(
-        File(exists=True),
-        mandatory=True,
-        desc="Denoised BOLD data.",
-    )
-    interpolated_filtered_bold = traits.List(
+    denoised_interpolated_bold = traits.List(
```
mattcieslak, line 269: nitpick: For these I've been using `InputMultiObject` instead of `traits.list`

xcp_d/interfaces/nilearn.py
```diff
@@ -391,24 +361,19 @@ def _run_interface(self, runtime):
...
-        uncensored_denoised_img.header.set_zooms(pixdim)
-        uncensored_denoised_img.to_filename(self._results["uncensored_denoised_bold"])
-
-        self._results["interpolated_filtered_bold"] = os.path.join(
+        self._results["denoised_interpolated_bold"] = os.path.join(
             runtime.cwd,
             "filtered_denoised.nii.gz",
         )
-        filtered_denoised_img = masker.inverse_transform(interpolated_filtered_bold)
+        filtered_denoised_img = masking.unmask(
```
mattcieslak, line 368: I didn't know about this function, very cool

mattcieslak (approved):
Is the --no-dcan-qc PR also going in before the next release?

docs/outputs.rst
tsalo: Despiking isn't represented in the design matrix, because it's done independently for each voxel and is only applied to the data (not the confounds).

docs/workflows.rst
tsalo: Justify the choice of `lstsq` over QR decomposition, or justify the decision to denoise the interpolated data?

mattcieslak: lstsq

xcp_d/interfaces/concatenation.py
mattcieslak: oh interesting! Now that I think of it, I don't think I use this anywhere where there might be only one input

tsalo:
The `--no-dcan-qc` option is in #1086. I need to merge it before the next minor release (0.7.0), and the make a follow-up PR that actually removes `--dcan-qc`.

docs/workflows.rst
tsalo: Gotcha. The reason we need to do it is because QR decomposition doesn't estimate betas, which we need to denoise the interpolated data. Should I basically just add that info to the documentation?

tsalo: Here is my updated text. WDYT?

```rst
XCP-D modifies Nilearn's approach in the following ways:

1. XCP-D uses :func:`numpy.linalg.lstsq` to estimate betas instead of QR decomposition,
   in order to denoise the interpolated data as well.
   -  QR decomposition will not produce betas that can be applied to the interpolated data.

2. XCP-D sets any leading or trailing high-motion volumes to the closest low-motion volume's values
   instead of extrapolating those volumes or removing them completely.

Both of these modifications allow XCP-D to produce a denoised, interpolated BOLD time series,
while Nilearn only produces the denoised, _censored_ BOLD time series.
The interpolated BOLD time series is necessary for DCAN-specific tools,
such as `biceps <https://biceps-cmdln.readthedocs.io/en/latest/>`_.
```

PR #1104, opened 2024-03-23
Add BIDSURI and GenerateMetadata interfaces to track Sources

tsalo:
It looks like GenerateMetadata's input_names aren't being set as attributes of the inputs field when called in a MapNode. I'll need to investigate.


tsalo:
The Sources going into the `meta_dict` in the concatenation workflow needs to be transposed.

PR #1135, opened 2024-04-08
Use standardized DVARS from Nipype

tsalo:
Closes none, but stems from [this NeuroStars post](https://neurostars.org/t/higher-dvars-in-xcp-d-than-fmriprep/29008).

- Replace XCP-D's DVARS formula with the one from Nipype/fMRIPrep.
    - I had to remove the intensity-normalization element because it doesn't seem to work appropriately on the denoised data. Basically, the intensity normalization divide by the median value across the brain and across time in the BOLD data, but that median value is very low (~0.01) in the denoised data, which results in infinite values in the denoised DVARS.
- Adopt the "standardized" DVARS from the new function for our QC measures and plots.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,6 +27,7 @@ dependencies = [
     "nilearn == 0.10.3",  # 0.10.2 raises error with compcor from fMRIPrep 23
     "nipype ~= 1.8.5",
     "niworkflows == 1.10.1",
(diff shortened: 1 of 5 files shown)


tsalo:
The current problem is that denoising results in very low median values (e.g., 0.01), which is used as the denominator when calculating standardized DVARS. The "big number divided by very small number" step leads to Infs in some cases.

PR #1152, opened 2024-04-23
Fix exit code bug in `cli.run.main()`

tsalo:
Closes none, but addresses a bug reported by @erikglee when running XCP-D on CBRAIN.

- Fix a bug in which I referenced an attribute of the Config object that doesn't exist. I copied the code from fMRIPrep without noticing the problem. The workflow still successfully ran, but the exit code was 1 (failure) instead of 0 (success) regardless.
- Improve integration tests by using `xcp_d.cli.run.main()` directly.

### DIFF
--- a/xcp_d/cli/run.py
+++ b/xcp_d/cli/run.py
@@ -153,14 +153,6 @@ def main():
                 f"boilerplate text found in {boiler_file}.",
             )
 
-        if config.workflow.run_reconall:
-            from niworkflows.utils.misc import _copy_any
-            from templateflow import api
-
-            dseg_tsv = str(api.get("fsaverage", suffix="dseg", extension=[".tsv"]))
-            _copy_any(dseg_tsv, str(config.execution.xcp_d_dir / "desc-aseg_dseg.tsv"))
-            _copy_any(dseg_tsv, str(config.execution.xcp_d_dir / "desc-aparcaseg_dseg.tsv"))
-
         errno = 0
 
     finally:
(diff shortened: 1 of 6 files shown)


xcp_d/cli/run.py
```diff
@@ -23,7 +23,7 @@ def main():
         write_dataset_description,
     )
 
-    parse_args()
+    parse_args(args=sys.argv[1:])
```
tsalo, line 26: Need to pass the arguments in directly so I can patch them in in the integration tests.

xcp_d/cli/run.py
```diff
@@ -153,14 +153,6 @@ def main():
                 f"boilerplate text found in {boiler_file}.",
             )
 
-        if config.workflow.run_reconall:
```
tsalo, line 156: This was the problem- `config.workflow.run_reconall` is an attribute that doesn't exist in the XCP-D Config, but its location within this try/except/else/finally statement meant that it wasn't doing anything except blocking the line that sets `errno` to 0.

xcp_d/tests/test_cli.py
```diff
@@ -359,27 +356,12 @@ def _run_and_generate(test_name, parameters, input_type):
...
-        subject_list=config.execution.participant_label,
-        output_dir=config.execution.xcp_d_dir,
-        run_uuid=config.execution.run_uuid,
-        config=str(load_data("reports-spec.yml")),
-        packagename="xcp_d",
-    )
+
+    argv = ["xcp-d"] + parameters
+    with patch.object(sys, "argv", argv):
+        run.main()
```
tsalo, line 364: So! Much! Simpler!!!!!!!

PR #1234, opened 2024-08-15
Make collect_data work for T1w+T2w nibabies derivatives

tsalo:
Closes none, but addresses a bug reported by @erikglee.

Here's the issue: With T1w+T2w derivatives in Nibabies, the `anat_dseg` file may have `space-T2w` instead of no space entity like we expect. 

Fixing this led us down the rabbit hole, as the T1w and T2w produced by Nibabies are in different spaces (native T1w and native T2w, respectively). XCP-D assumes that they would be in the same space, so the same anatomical-to-template transform could be applied to both. Unfortunately, that's not the case with Nibabies derivatives. Also, there is no T1w-to-T2w transform that we could use to get them in the same space (not that I'd want to do that, since it would be an extra transform to apply).

The only things this should affect are the executive summary brainsprite and figures where templates are overlaid on top of the anatomical images.

### DIFF
--- a/xcp_d/data/io_spec.yaml
+++ b/xcp_d/data/io_spec.yaml
@@ -11,7 +11,10 @@ queries:
             datatype: anat
             desc: null
             extension: .nii.gz
-            space: null
+            space:
+            - T1w
+            - T2w
+            - null
             suffix: dseg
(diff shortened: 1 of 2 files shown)


tsalo:
I'll look into testing later. Going to merge now so @erikglee can test ASAP.

PR #1249, opened 2024-09-05
Fix subject sphere query

tsalo:
Closes none, but addresses a bug discovered by Julia Moser.
Apparently XCP-D was including "space": None in the query, which is not right for finding the sphere files.

- Fix subject sphere query


tsalo:
This should only affect derivatives from very new versions of Nibabies and fMRIPrep (~24.0.0).

PR #1255, opened 2024-09-11
Replace restricted confound sets with config files

### LINKED ISSUE #1039: Add `--derivatives` parameter to command-line interface
## Summary

fMRIPrep and ASLPrep now support a `--derivatives` parameter, where users can provide paths to pre-computed derivatives.
This could be useful if we want to provide paths to FreeSurfer/MCRIBS derivatives, or maybe even other postprocessing pipelines?

### LINKED ISSUE #1074: Support ANATICOR?
## Summary

Related to https://github.com/nipreps/fmriprep/issues/1457. Relevant papers: https://www.hindawi.com/journals/jam/2013/935154/ and http://dx.doi.org/10.1016/j.neuroimage.2010.04.246.

I doubt we want to actually create the ANATICOR regressors, but we could conceivably support it as a nuisance regression option. The ANATICOR regressors could be created by an fMRIPost pipeline.

### LINKED ISSUE #1231: Add "27P" and "36P" confound regression options without GSR
## Summary
It would be great to have the option to run "27P" and "36P" confound regression *without* global signal regressors. I.e.:
- `27P-nogsr` or `26P`: 24 motion parameters and derivatives + WM + CSF 
- `36P-nogsr` or `32P`: 24 motion parameters and derivatives + 4 WM regressors + 4 CSF regressors

## Additional details
- Benefit for practice: Use when GSR is not wanted or to evaluate the effect of global signal regression
- Consistency within confound regressor options: There are options to run aCompCor and AROMA with and without GSR. Adding such an option to the other regressor sets would feel sensible.

## Next steps
I admit that I did not look deep into your code. However, adding...

```python
# Get rot and trans values, as well as derivatives and square, WM and CSF.
"27P-nogsr": {
    "strategy": ["motion", "wm_csf"],
    "motion": "full",
    "wm_csf": "basic",
},
# Get rot and trans values, as well as derivatives and square, WM, and CSF.
# Add the square and derivative of the WM and CSF.
"36P-nogsr": {
    "strategy": ["motion", "wm_csf"],
    "motion": "full",
    "wm_csf": "full",
},
```

...to [xcp_d/utils/confounds: load_confound_matrix()](https://github.com/PennLINC/xcp_d/blob/main/xcp_d/utils/confounds.py#L177) plus some higher-level adjustments could suffice?

Thanks and thank you a lot for your great work!

### LINKED ISSUE #1252: Allow users to define denoising strategies with configuration files
## Summary

This is something I'm already working on in response to #1231. I just opened this issue to track my progress because I open a PR.

## Additional details

This will allow users to combine confounds from different pipelines, such as fMRIPrep and fMRIPost-AROMA. It should also reduce the work needed to build custom confounds.

## Next steps

1. Convert builtin nuisance regression strategies (e.g., 36P) to YAML files.
    - Considerations to take into account:
         - Definition of signal regressors.
         - Support regular expressions.
         - Apply transforms to confounds (e.g., run PCA on aggregated transforms). (long-term goal)
         - Allow voxel-wise regressors.
         - Automatically determine which (if any) columns should be filtered if users want to apply a motion filter.
2. Modify the CLI to accept either a builtin strategy name or a path to a YAML file.
3. Create a `--derivatives` CLI argument with labeled BIDS datasets.
4. Load confound _files_ from the datasets based on the config.
5. Combine the loaded confounds.
    - For tabular confounds, load the files, select the requested columns, and combine them into a single DataFrame.
    - For voxel-wise confounds, check the space, resolution, affine, etc. to ensure that they all match.
6. For the confound regression, use the combined confound DataFrame and the list of voxel-wise confounds.
7. Use the YAML file's other data to build the boilerplate and metadata.
    - I'm just going to assume that any references are already in XCP-D's bib file for now.
9. Document how to create minimal BIDS datasets with custom confounds.

tsalo:
Closes #1252, closes #1231, closes #1039, closes #1074.

Continues fixing Freesurfer issues from #1273.

## Changes to inputs

- Allow users to provide their own confounds configurations via YAML files pass to the `--nuisance-parameters` parameter.
- Add `--derivatives` parameter.
    - The primary derivatives dataset (i.e., the one provided through a positional argument instead of the `--derivatives` keyword argument) is internally referred to as "preprocessing".
    - Derivatives are linked to the confound configuration files via keywords. E.g., `--derivatives aroma=/path/to/fmripost-aroma`.
- Remove `--custom-confounds` parameter. **THIS IS A BREAKING CHANGE**

## Changes to outputs

- The motion file will not have `desc-filtered` in the filename anymore. **THIS IS A BREAKING CHANGE**
- The motion file will have both the original and filtered versions of the motion parameters (plus RMSD and FD) instead of just the filtered versions.
- The motion parameters in any design matrix files will, if filtered, have a `_filtered` suffix in the column name.

## Internal changes

- Separate handling of the motion parameters used for outlier detection from motion parameters that are used in the nuisance regression.


tsalo:
I tested out `aroma_gsr` using fMRIPrep and fMRIPost-AROMA derivatives, and it worked! Still some bugs though:

1. The top-level `Sources` field of the design.json file is wrong. It has the fMRIPrep confounds file, without a BIDS-URI.
2. The column-level `Sources` fields of the design.json file are wrong. They only point to the fMRIPost-AROMA confounds file, even for columns pulled from the fMRIPrep confounds file. They're BIDS-URIs at least.
3. The motion.json file includes metadata from all of the columns in the fMRIPrep confounds json, instead of just the motion columns. The top-level `Sources` field is correct though.

PR #1256, opened 2024-09-11
Remove n_procs for nodes that do not use multiple cores

tsalo:
Closes none, but addresses an issue identified by @mattcieslak.

- For any single-threaded nodes in any workflow, drop `n_procs=omp_nthreads`.
- Also drop `omp_nthreads=*` from any nodes using that. Pretty sure it doesn't do anything. 
- Create a WBCommand base interface that sets OMP_NUM_THREADS.
- Use the MultiProc plugin for integration tests, which speeds things up by a fair amount.

### DIFF
--- a/xcp_d/workflows/anatomical/surface.py
+++ b/xcp_d/workflows/anatomical/surface.py
@@ -527,9 +527,10 @@ def init_generate_hcp_surfaces_wf(name="generate_hcp_surfaces_wf"):
     )
 
     generate_midthickness = pe.Node(
-        SurfaceAverage(),
+        SurfaceAverage(num_threads=config.nipype.omp_nthreads),
         name="generate_midthickness",
         mem_gb=2,
+        n_procs=config.nipype.omp_nthreads,
     )
     workflow.connect([
         (inputnode, generate_midthickness, [
@@ -560,8 +561,12 @@ def init_generate_hcp_surfaces_wf(name="generate_hcp_surfaces_wf"):
 
     # Generate (very-)inflated surface from standard-space midthickness surface.
     inflate_surface = pe.Node(
-        SurfaceGenerateInflated(iterations_scale_value=0.75),
+        SurfaceGenerateInflated(
+            iterations_scale_value=0.75,
+            num_threads=config.nipype.omp_nthreads,
+        ),
         mem_gb=2,
+        n_procs=config.nipype.omp_nthreads,
         name="inflate_surface",
     )
     workflow.connect([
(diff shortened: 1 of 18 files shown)


xcp_d/workflows/anatomical/surface.py
```diff
@@ -800,13 +784,11 @@ def init_ants_xfm_to_fsl_wf(mem_gb, omp_nthreads, name="ants_xfm_to_fsl_wf"):
         BinaryMath(expression="img * -1"),
```
mattcieslak, line 784: Do you know how wb_command handles concurrency?

xcp_d/workflows/anatomical/volume.py
```diff
@@ -168,7 +167,6 @@ def init_postprocess_anat_wf(
                 ),
                 name="warp_t1w_to_template",
                 mem_gb=2,
-                n_procs=omp_nthreads,
```
mattcieslak, line 171: I would keep this and set ``num_threads=omp_nthreads`` in the ApplyTransforms in the interface

xcp_d/workflows/anatomical/volume.py
```diff
@@ -189,7 +187,6 @@ def init_postprocess_anat_wf(
                 ),
                 name="warp_t2w_to_template",
                 mem_gb=2,
-                n_procs=omp_nthreads,
```
mattcieslak, line 192: Same here

xcp_d/workflows/bold/metrics.py
```diff
@@ -145,7 +145,6 @@ def init_alff_wf(
         ComputeALFF(TR=TR, low_pass=low_pass, high_pass=high_pass),
         mem_gb=mem_gb["resampled"],
```
mattcieslak, line 146: I hope the memory limits will be enough to prevent nipype from running too many of these at once

mattcieslak (changes requested):

xcp_d/workflows/anatomical/surface.py
tsalo: I need to read this tomorrow, but here's the relevant documentation:

>wb_command -parallel-(salo) Taylors-MBP:code taylor$ wb_command -parallel-help
   Many processing commands make use of multithreading so that scripts can
   finish more quickly.  By default, these commands will use all available
   cores on the system.  Because we use OpenMP for the parallelization, this
   behavior can be controlled through various environment variables.  In
   particular, 'OMP_NUM_THREADS' will set the maximum number of threads it will
   use, when exported as an environment variable.  Additionally, some shells,
   such as bash, have syntax that allows you to set an environment variable for
   a single command or script, for instance:
>
> $ OMP_NUM_THREADS=4 wb_command -volume-smoothing input.nii.gz 4 output.nii.gz
>
>   If you have a multi-socket system, be aware that the parallelization can be
   much slower when threads are on different sockets, and this interacts badly
   with the default behavior of using all available cores.  It is advisable to
   use other tools to restrict the entire script to execute on a single socket,
   especially if a queueing system is involved.
>
>   Also note that wb_view contains a few features that use multithreading
   (dynamic connectivity, border optimize), which can be controlled by setting
   the same environment variables before launching wb_view.

mattcieslak: This might be built into the interfaces already. Freesurfer and afni also work using this variable

mattcieslak: It's not. The wbcommand class in nipype doesn't have multithreading set up in it

tsalo: I can try creating a version that inherits from the original.

mattcieslak: that might be worth the trouble for some of these

tsalo:
The integration tests don't seem to contribute to coverage now that we're using the MultiProc plugin, but I think the speed-up is worth the drop in coverage, so I'm going to merge this.

PR #1258, opened 2024-09-12
Parallelize ALFF

### LINKED ISSUE #1257: Parallelize ALFF and denoising interfaces
## Summary

Cool idea from @mattcieslak.

Both interfaces should work on individual vertices/voxels' time series.

tsalo:
Closes #1257. I don't think it makes sense to try parallelizing the denoising interface until I have voxel-wise regressors up and running. And even then I would only want to parallelize _when_ there are voxel-wise regressors.

- Use `ProcessPoolExecutor` to parallelize ALFF.


tsalo:
Wow that _really_ didn't work.

PR #1278, opened 2024-10-02
replacing iirnotch() with butter()

### LINKED ISSUE #1272: Possible bug in DCANBoldProc that may or may not be relevant to XCP-D
@cindyhfls noticed a possible bug in our code base for DCANBoldProc. Our team is going to check it out soon. I wanted to share in case the issue is present in XCP-D as well! 

See: https://github.com/DCAN-Labs/dcan_bold_processing/issues/28

cindyhfls:
Closes #1272.

The iirnotch() from scipy was second-order, and filtfilt() doubles it. So if repeated application is used here, n_filter_applications should be int(np.floor(motion_filter_order/4)) not int(np.floor(motion_filter_order/2)).

notch filter removes the center frequency much more than the other frequencies in the band but butterworth filter is more flat and removes the whole band more evenly

Also applying the filter multiple times instead of changing the order in the filter design seems to be reducing the signal amplitude.
For low-order like 4, it barely matters but if the order is increased to 10, the effect of repeated filtering versus changing the order in the butter() function seems to be more profound. I don't know why it would behave like that but this is just an observation.
image

### DIFF
--- a/xcp_d/utils/confounds.py
+++ b/xcp_d/utils/confounds.py
@@ -405,19 +405,22 @@ def motion_regression_filter(
         bandwidth = np.abs(np.diff(stopband_hz))
 
         # Create filter coefficients.
-        b, a = iirnotch(freq_to_remove, freq_to_remove / bandwidth, fs=sampling_frequency)
-        n_filter_applications = int(np.floor(motion_filter_order / 2))
+        b, a = butter(
+            motion_filter_order / 2,
+            stopband_hz,
+            btype="bandstop",
+            output="ba",
+            fs=sampling_frequency,
+        )
 
         filtered_data = data.copy()
-        for _ in range(n_filter_applications):
-            filtered_data = filtfilt(
-                b,
-                a,
-                filtered_data,
-                axis=0,
-                padtype="constant",
-                padlen=data.shape[0] - 1,
-            )
+        filtered_data = filtfilt(
+            b,
+            a,
+            data,
+            axis=0,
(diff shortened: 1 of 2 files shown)


xcp_d/utils/confounds.py
```diff
@@ -405,19 +405,22 @@ def motion_regression_filter(
...
-                padtype="constant",
-                padlen=data.shape[0] - 1,
-            )
+        filtered_data = filtfilt(
+            b,
+            a,
+            data,
+            axis=0,
+            padtype="constant",
+            padlen=data.shape[0] - 1,
```
tsalo, line 423:
```suggestion
            padlen=data.shape[0] - 1,
        )
```

tsalo:
Thanks for opening this. Can you split the HCP-D ingression into a separate PR though?

cindyhfls:
that was for my personal use, I did not mean to include that but I don't know how to make it a separate commit. ==

LuciMoore:
@tsalo it's my understanding that, for DCAN Bold Proc, we plan on integrating the portion of the PR that changes the filter order, but will not be replacing the IIRNotch filter with Butterworth. So I would recommend doing the same for XCP-D. @madisoth can you please confirm?

tsalo:
@cindyhfls when do you think you'll be able to update the PR?

cindyhfls:
Sorry I'm not very familiar with editing existing PR to make changes. I hope it's resolved?

PR #1279, opened 2024-10-04
Hcp make wm vent mask

cindyhfls:
Making individual masks for HCP output followed from this: https://neurostars.org/t/differences-in-dcan-derivatives-and-hcp-ya-outputs/30631/2

### DIFF
--- a/xcp_d/ingression/hcpya.py
+++ b/xcp_d/ingression/hcpya.py
@@ -23,6 +23,7 @@ from xcp_d.ingression.utils import (
     copy_files_in_dict,
     plot_bbreg,
     write_json,
+    make_masks,
 )
 from xcp_d.utils.filemanip import ensure_list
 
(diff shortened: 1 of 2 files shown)


cindyhfls:
This needs the installation of fsl: https://nipype.readthedocs.io/en/latest/users/examples/fmri_fsl.html

tsalo:
I recently managed to remove FSL from the XCP-D Docker image and I'm not planning to add it back in. As far as I can tell, you can accomplish everything you're currently doing with fslmaths with Python tools. I would recommend using Python libraries, as the other ingression functions do.

cindyhfls:
i tried that but I could not reproduce the same erosion effect with python, feel free to implement thatOn Oct 15, 2024, at 8:14 AM, Taylor Salo ***@***.***> wrote:﻿
I recently managed to remove FSL from the XCP-D Docker image and I'm not planning to add it back in. As far as I can tell, you can accomplish everything you're currently doing with fslmaths with Python tools. I would recommend using Python libraries, as the other ingression functions do.

PR #1283, opened 2024-10-07
Track Sources with BIDSURI interface

tsalo:
Takes over for #1104.


tsalo:
Looks like I still need to transpose the list of lists of Sources in the concatenation workflow.

PR #1284, opened 2024-10-08
Fix NIfTI-surface connection

tsalo:
Closes none, but addresses an issue reported in https://neurostars.org/t/xcp-d-some-connections-were-not-found/30667.

This bug arises when users (1) provide derivatives with mesh files, (2) enable abcc-qc, and (3) enable NIfTI processing. Because the mesh files are available and abcc-qc is enabled, the surface-processing workflow is enabled (though without normalization to fsLR because `--warp-surfaces-native2std` isn't enabled), but the BOLD post-processing workflow doesn't use the post-processed surfaces.

This bug should affect versions 0.8.0 - 0.10.rc1, although it _might_ affect earlier versions.

- Only pass midthickness files to the BOLD post-processing workflow if CIFTI processing is enabled.

### DIFF
--- a/xcp_d/tests/data/pnc_cifti_filter.json
+++ b/xcp_d/tests/data/pnc_cifti_filter.json
@@ -1,6 +1,8 @@
 {
     "bold": {
         "task": ["rest"],
-        "run": [null]
+        "run": [null],
+        "space": "fsLR",
+        "den": "91k"
     }
 }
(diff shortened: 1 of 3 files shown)


tsalo:
Bug successfully reproduced:

```
E               Exception: Some connections were not found
E               Module postprocess_0_wf has no input called inputnode.lh_midthickness
E               
E               Module postprocess_0_wf has no input called inputnode.rh_midthickness
```

PR #1300, opened 2024-10-21
Apply notch filter appropriate number of times based on `--motion-filter-order`

### LINKED ISSUE #1272: Possible bug in DCANBoldProc that may or may not be relevant to XCP-D

tsalo:
Closes #1272 and closes #1278. Thanks to @cindyhfls for detecting the problem and figuring out the solution.

- Divide the `--motion-filter-order` parameter by 4 instead of 2 when using the "notch" motion filter, as `iirnotch` is a second-order filter and `filtfilt` applies the filter both forwards and backwards (equaling 4 total applications).
- Update the documentation to reflect this.
- Fix a surprising bug related to exact-time correlation matrices. Not sure why it never came up before. See [8f8816f](https://github.com/PennLINC/xcp_d/pull/1300/commits/8f8816f0d3fd0a6ba37777328df5bf9dee70ec11).

### DIFF
--- a/xcp_d/cli/parser.py
+++ b/xcp_d/cli/parser.py
@@ -373,7 +373,15 @@ This parameter is used in conjunction with ``motion-filter-order`` and ``band-st
         dest="motion_filter_order",
         default=4,
         type=int,
-        help="Number of filter coeffecients for the motion parameter filter.",
+        help=(
+            "Number of filter coefficients for the motion parameter filter. "
+            "If the motion filter type is 'lp', the order is divided by 2 as filtfilt applies "
+            "the filter twice. "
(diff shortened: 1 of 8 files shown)


tsalo:
@cindyhfls would you mind reviewing my changes? Also, given that you detected the problem and proposed the fix, would you like me to add you to our CITATION.cff file? If so, can you provide your information to match this?

https://github.com/PennLINC/xcp_d/blob/16b16f72817b79b5c98382e5c4e4abd5da1be183/CITATION.cff#L13-L17

PR #1326, opened 2024-11-18
Write out correlations when `--create-matrices` doesn't include "all"

tsalo:
Closes none, but addresses a bug identified in https://neurostars.org/t/filenotfounderror-no-bold-data-found-in-allowed-spaces-fslr/30968.

- Set `config.execution.output_correlations` to True when exact scans are provided through `--create-matrices`, but "all" is not included.


tsalo:
It looks like this might require a bit more thought.

PR #1363, opened 2025-01-21
Support multi-step composite transforms from Nibabies (patch)

### LINKED ISSUE #1348: New format for Nibabies anat-to-MNI transforms does not work with surface-warping workflow
## Summary

@erikglee came across this bug before the break. It looks like the new anat-to-MNI transforms being generated by Nibabies (as of https://github.com/nipreps/nibabies/pull/424) are not composite transforms that the surface-warping workflow can use. 

## Additional details

- xcp_d version: 0.10.1
- Docker version:
- Apptainer version:

### What were you trying to do?

Process recently-processed Nibabies derivatives using XCP-D with `--warp-surfaces-native2std y` enabled.

### What did you expect to happen?

XCP-D should be able to use the transforms generated by Nibabies to warp meshes to fsLR (aligned to MNI152NLin6Asym).

### What actually happened?

```
Traceback (most recent call last):
  File "/usr/local/miniconda/lib/python3.10/site-packages/nipype/pipeline/plugins/multiproc.py", line 67, in run_node
    result["result"] = node.run(updatehash=updatehash)
  File "/usr/local/miniconda/lib/python3.10/site-packages/nipype/pipeline/engine/nodes.py", line 527, in run
    result = self._run_interface(execute=True)
  File "/usr/local/miniconda/lib/python3.10/site-packages/nipype/pipeline/engine/nodes.py", line 645, in _run_interface
    return self._run_command(execute)
  File "/usr/local/miniconda/lib/python3.10/site-packages/nipype/pipeline/engine/nodes.py", line 771, in _run_command
    raise NodeExecutionError(msg)
nipype.pipeline.engine.nodes.NodeExecutionError: Exception raised while executing Node disassemble_h5.

Cmdline:
	CompositeTransformUtil --disassemble /data/sub-X/ses-Y/anat/sub-X_ses-Y_run-1_from-T2w_to-MNI152NLin6Asym_mode-image_xfm.h5 T1w_to_MNI152NLin6Asym
Stdout:
	Transform File /data/sub-X/ses-Y/anat/sub-X_ses-Y_run-1_from-T2w_to-MNI152NLin6Asym_mode-image_xfm.h5 is not a Composite Transform.
Stderr:

Traceback:
	RuntimeError: subprocess exited with code 1.
```

## Reproducing the bug

Preprocess fMRI data with Nibabies after https://github.com/nipreps/nibabies/pull/424 was merged with `--cifti-output` and Freesurfer, then run XCP-D with `--file-format cifti` and `--warp-surfaces-native2std y`.

tsalo:
Closes #1348.

- Disassemble composite transforms more flexibly, allowing multiple affines and warps instead of hardcoding just one of each.
- Loop over affines and warps from the composite transforms, applying them to the surfaces sequentially.


xcp_d/workflows/anatomical/surface.py
```diff
@@ -995,35 +1033,36 @@ def init_warp_one_hemisphere_wf(
     # Apply FLIRT-format anatomical-to-template affine transform to 32k surfs
     # I think this makes it so you can overlay the pial and white matter surfaces on the
     # associated volumetric template (e.g., for XCP-D's brainsprite).
-    apply_affine_to_fsLR32k = pe.MapNode(
+    # TODO: Need to loop over affine-warp pairs here rather than apply them in parallel.
+    apply_affine_to_template = pe.MapNode(
```
tsalo, line 1037: This is the tough section. We have two separate nodes: apply_affine and apply_warpfield, which make sense when there's one of each. However, in the scenario where we have _two_ of each, I'm not sure how to loop over the affine-warpfield pairs and apply these steps twice.

One workaround (which I hate) would be to infer the number of affine-warp pairs in the composite transform outside of the nipype workflow (i.e., using regular Python code in the workflow function) and use that to make a loop using `niu.Select`.

tsalo:
To make this work, I think I need to do the following:

1. Track the order of the transforms coming out of the composite transforms.
    - Will need to modify the transform disassembly interface to output this.
2. Match up the warp and inverse warp transforms so that they're applied together.
    - At the moment, they're probably aligned in reverse order (e.g., anat-to-MNIInfant is used in conjunction with MNI152-to-MNIInfant instead of MNIInfant-to-anat).

PR #1368, opened 2025-01-23
Use trimesh to generate brainsprite

### LINKED ISSUE #943: Remove hardcoded crosshairs from brainsprite
## Summary

It would be great to remove the green and blue lines/planes from the brainsprite (see below).

image

## Additional details

Improved visualization.

## Next steps

I think this will require modification of the workbench scene file(s), which I do not have the expertise to do.

### LINKED ISSUE #944: Set entire background of brainsprite to black
## Summary

Currently, the background of the brainsprite outside of the FOV is white. We'd like it to be black.

image

## Additional details

Better visualization.

## Next steps

This will probably require modifying the workbench scene file, which I don't know how to do.

tsalo:
Closes #943 and closes #944.

- Replace wb_command-based scenes with trimesh figures when building the mosaic for the brainsprite. This has the side benefits of (1) speeding up the workflow, (2) removing the crosshairs from the briansprite, and (3) setting the full background to black.


mattcieslak (approved):
big improvement!

PR #1369, opened 2025-01-24
Stop autoescaping the boilerplate

mattcieslak:
Currently the boilerplate, markdown and latex in the executive summary are being rendered as plain text. This should revert them to be rendered as html.


tsalo (approved):
Awesome!

PR #1379, opened 2025-02-05
Add session-id parameter

### LINKED ISSUE #1377: Add `--session-id` argument
## Summary

This would feed into the BIDS filter files.

## Additional details

This is distinct from supporting session-wise anatomical data.

tsalo:
Closes #1377.

- Add `--session-id` parameter to CLI. The `session_id` variable is passed along into all of the BIDS filters. A value of `Query.NONE` is also passed along, primarily for anatomical data that might not be at the session level.


tsalo:
I should update the CLI parsing tests too.

mattcieslak (approved):
I can't think of any reasons why this wouldn't work, and you added a test where it's used, so I say this is good to go

PR #1381, opened 2025-02-06
Only estimate head radius if necessary

tsalo:
Closes none. This is just a minor optimization that eliminates a step that warps the standard-space anatomical brain mask to anatomical space when it's not necessary.

- Don't estimate the head radius if `--head-radius` isn't "auto".

### DIFF
--- a/xcp_d/workflows/base.py
+++ b/xcp_d/workflows/base.py
@@ -361,26 +361,28 @@ It is released under the [CC0](https://creativecommons.org/publicdomain/zero/1.0
                 ])  # fmt:skip
 
     # Estimate head radius, if necessary
-    # Need to warp the standard-space brain mask to the anatomical space to estimate head radius
-    warp_brainmask = ApplyTransforms(
-        input_image=subj_data['anat_brainmask'],
-        transforms=[subj_data['template_to_anat_xfm']],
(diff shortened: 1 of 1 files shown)


xcp_d/workflows/base.py
```diff
@@ -361,26 +361,28 @@ def init_single_subject_wf(subject_id: str):
...
+        warp_brainmask = ApplyTransforms(
+            input_image=subj_data['anat_brainmask'],
+            transforms=[subj_data['template_to_anat_xfm']],
+            reference_image=subj_data[anat_mod],
+            num_threads=2,
+            interpolation='GenericLabel',
+            input_image_type=3,
+            dimension=3,
+        )
+        os.makedirs(config.execution.work_dir / workflow.fullname, exist_ok=True)
```
mattcieslak, line 376: Why does this directory need to be explicitly made?

mattcieslak: ohhh I see, because this is being run during the workflow build, not as part of the workflow itself

mattcieslak (approved):
I'm always reluctant to add an actual computation to the workflow building part, but I'm not sure how you can get around it here. This seems fine overall though

PR #1387, opened 2025-02-07
Parallelize image denoising

mattcieslak:
The image denoising function can probably be parallelized to run in chunks, such that you can keep all the available cpus running at all times. This is an attempt to use multiprocessing to split up input data to be processed on all cpus separately and ultimately combined

### DIFF
--- a/xcp_d/interfaces/nilearn.py
+++ b/xcp_d/interfaces/nilearn.py
@@ -264,6 +264,7 @@ class _DenoiseImageInputSpec(BaseInterfaceInputSpec):
     low_pass = traits.Float(mandatory=True, desc='Lowpass filter in Hz')
     high_pass = traits.Float(mandatory=True, desc='Highpass filter in Hz')
     filter_order = traits.Int(mandatory=True, desc='Filter order')
+    num_threads = traits.Int(1, usedefault=True, desc='denoise on this many cpus')
 
 
 class _DenoiseImageOutputSpec(TraitedSpec):
@@ -335,6 +336,7 @@ class DenoiseCifti(NilearnBaseInterface, SimpleInterface):
             high_pass=high_pass,
             filter_order=self.inputs.filter_order,
             TR=self.inputs.TR,
+            num_threads=self.inputs.num_threads,
         )
 
         # Transpose from TxS (nilearn order) to SxT (xcpd order)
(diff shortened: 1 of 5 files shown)


tsalo:
I think the slice plotting might be kind of slow due to the mosaic-building node, which loads all of the individual figures into memory. I have no clue about the coverage plotting step though.

xcp_d/workflows/parcellation.py
```diff
@@ -399,15 +399,15 @@ def init_parcellate_cifti_wf(
     ])  # fmt:skip
 
     # Threshold node coverage values based on coverage threshold.
+    # Id doesn't benefit from multiple cpus
```
tsalo, line 402:
```suggestion
    # It doesn't benefit from multiple cpus
```

xcp_d/workflows/bold/connectivity.py
```diff
@@ -355,7 +355,6 @@ def init_functional_connectivity_cifti_wf(mem_gb, exact_scans, name='connectivit
                 vmax=1,
             ),
             name='plot_coverage',
-            mem_gb=mem_gb['resampled'],
```
tsalo, line 358: Why drop these?

xcp_d/workflows/anatomical/plotting.py
```diff
@@ -137,19 +137,10 @@ def init_brainsprite_figures_wf(
...
-                input_names=[
-                    'nifti',
-                    'lh_wm',
-                    'rh_wm',
-                    'lh_pial',
-                    'rh_pial',
-                ],
-                output_names=['out_files'],
-            ),
+            PlotSlicesForBrainSprite(n_procs=config.nipype.omp_nthreads),
```
tsalo, line 140: I checked the brainsprites from the integration tests and they looked good, so I don't think there's an issue with multiprocessing shuffling the slices.

xcp_d/interfaces/execsummary.py
```diff
@@ -465,6 +471,31 @@ def _run_interface(self, runtime):
...
+    rh_pial : trimesh.Trimesh
+        Right hemisphere pial surface loaded as a Trimesh
+    lh_pial : trimesh.Trimesh
+        Left hemisphere pial surface loaded as a Trimesh
+    rh_wm : trimesh.Trimesh
+        Right hemisphere wm surface loaded as a Trimesh
+    lh_wm : trimesh.Trimesh
+        Left hemisphere wm surface loaded as a Trimesh
+    root_dir : str
+        Strin representing the directory where files will be written
```
tsalo, line 491:
```suggestion
        String representing the directory where files will be written
```

xcp_d/workflows/bold/connectivity.py
mattcieslak: We can leave this mystery for another day

tsalo (approved):
LGTM. Can write tests in separate PR.

PR #1388, opened 2025-02-07
Support `--abcc-qc` with `--file-format nifti`

tsalo:
Closes none, but addresses bug detected by user in https://neurostars.org/t/issue-with-running-xcp-d-with-nifti-file-format/31702.

- Hardcode the atlas format in `init_parcellate_surfaces_wf` as "cifti", so that it uses CIFTI atlases to parcellate surfaces.
- I also tested it out in ds001419_nifti test and it worked. I reverted that change since creating the brainsprite takes a while.


mattcieslak (approved):
LGTM

PR #1390, opened 2025-02-07
Add FAQ page to docs

tsalo:

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -40,3 +40,13 @@ While there are always new methods being developed, we have limited *XCP-D* to m
 have been validated and are commonly used in the field.
 If there is a method you would like to see included in *XCP-D*, please open an issue on GitHub,
 but please be aware that we may not be willing to include all methods.
+
+
+*******************************************************
+Can I use XCP-D on my data if it is not in BIDS format?
+*******************************************************
+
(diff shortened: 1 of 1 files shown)


singlesp:
@tsalo I want to probably move this rare edge case to faq rather than have it be the second thing a user sees after getting to the section on how to run xcp-d with apptainer:


Home Directory Handling in Apptainer
================================================================

When running XCP-D through Apptainer, the handling of home directories requires special attention. Here's what you need to know:

Default Behavior
--------------
By default, Apptainer will:
1. Mount your host system's home directory into the container at ``/home/$USER``
2. Set the container's ``$HOME`` environment variable to this location

Potential Issues
--------------
This default behavior can cause problems on some systems, particularly in HPC environments where home directory permissions or configurations may be restricted.

Workaround
--------------
If you encounter home directory-related issues, you can explicitly control the home directory handling with these Apptainer arguments:

.. code-block:: bash

   apptainer run \
      -B $HOME \ # Mount host's home directory to your $HOME in container
      --home $HOME \   # Set container's $HOME to your project directory
      --cleanenv xcp_d-<version>.simg \
      <xcp_d arguments>

.. note::
   Check with your system administrator if you're unsure about your system's Apptainer configuration or if you encounter permission issues.

PR #1394, opened 2025-02-07
Clarify mounting in apptainer instructions

singlesp:

`usage.rst`


docs/usage.rst
```diff
@@ -327,7 +327,7 @@ A Docker container can be created using the following command:
 
    docker run --rm -it \
       -v /path/to/fmriprep_dir:/fmriprep:ro \
-      -v /tmp/wkdir:/work:rw \
+      -v /path/to/wkdir:/work:rw \
```
mattcieslak, line 330: I thought we were not going to do any path aliasing in our examples?

mattcieslak (changes requested):
I think we should mount the directories to their original names in the docker example too. Otherwise this looks great

singlesp:
maybe move docker section to somewhere after apptainer?

docs/usage.rst
```diff
@@ -345,42 +317,60 @@ If you are computing on an :abbr:`HPC (High-Performance Computing)`, we recommen
...
+execution.
+
+A Docker container can be created using the following command:
+
+.. code-block:: bash
+
+   docker run --rm -it \ # docker args
+      -v /home/user/data/fmriprep \
+      -v /home/user/data/wkdir \
+      -v //home/user/data/xcpd_output \
```
mattcieslak, line 367:
```suggestion
      -v /home/user/data/xcpd_output \
```

mattcieslak (approved):
Approved, but one minor typo needs to be fixed

PR #1397, opened 2025-02-10
Note that automatic head radius estimation gives higher values than 50

### LINKED ISSUE #1292: Brain radius automatic estimation
I think using the brain mask gives a radius estimation slightly larger than what is used by people

I tried to calculate the radius using the formula provided and the brain mask from Freesurfer, and I got 61 mm for a BCP subject at 17 months and 76 mm for an HCP-D subject at 20 years. Both are higher than the 50 mm used as defaults for adults.

tsalo:
Closes #1292.


mattcieslak (approved):
looks good. Is this an option that warrants an alarm box? I'd guess no and that this is sufficient

tsalo:
I agree that this is sufficient.

PR #1411, opened 2025-02-24
Add nifti brain mask to hcpya ingression

smeisler:
Addresses https://neurostars.org/t/no-volumetric-outputs-when-running-xcp-d-on-hcp-ya-data/31905/3


tsalo (approved):
I was planning to hold off on approving until we tested the change, but it looks like the user tested it and it worked. Thanks for addressing this!

PR #1417, opened 2025-03-20
Use `mem_mb` instead of `mem_gb`

smeisler:
Closes none, but discussion on slack channel.

Use `mem_mb` in command line instead of gigabytes.

More consistent with other apps, allows the slurm memory environment variable to be directly input, which is helpful for BABS.


smeisler:
I thought the easiest route would be to change the CLI argument and then update from mb to gb in config so everything else can stay the same.

tsalo:
It looks like fMRIPrep [does it all in the CLI](https://github.com/nipreps/fmriprep/blob/2106034932c2bad27744daf6e4fe7053706c873e/fmriprep/cli/parser.py#L274-L283):

```python
    g_perfm.add_argument(
        '--mem',
        '--mem_mb',
        '--mem-mb',
        dest='memory_gb',
        action='store',
        type=_to_gb,
        metavar='MEMORY_MB',
        help='Upper bound memory limit for fMRIPrep processes',
    )
```

smeisler:
Changed it to be more fmriprep-like

tsalo:
I'm planning to make a patch release (0.10.7) with changes the DCAN team need, and then I can merge this PR before a minor release (0.11.0).

PR #1437, opened 2025-04-17
Collect and apply transforms between native anatomical spaces

### LINKED ISSUE #1431: Apply T2w-to-T1w transform as part of T2w normalization
## Summary

At least with Nibabies, the volumetric anatomical postprocessing warps the T1w to the template space for the executive summary figures, but the T2w is being warped with the same transform, which isn't correct for the T2w image. Nibabies outputs a T2w-to-T1w transform that can be collected and applied.

tsalo:
Closes #1431.

- When both T1w and T2w images are available and T1w-to-template transforms exist, but T1w-space T2w does not, then try to collect T2w-to-T1w transform and apply it to the T2w to get it into the same space as the T1w before normalizing.
- When both T1w and T2w images are available and _T2w-to-template_ transforms exist, but T2w-space T1w does not, then try to collect T1w-to-T2w transform and apply it to the T1w to get it into the same space as the T2w before normalizing.


xcp_d/data/tests/skeletons/nibabies_ignore_t1w.yml
mattcieslak, line 1: <3 skeletons

xcp_d/tests/test_utils_bids_collect_data.py
mattcieslak, line 1: This is a great approach to testing these kinds of functions

xcp_d/utils/bids.py
```diff
@@ -272,22 +272,22 @@ def collect_data(
...
+            t2wspace_t1w_found = bool(
                 layout.get(
                     return_type='file',
                     subject=participant_label,
                     **queries['t1w'],
                 )
             )
-            if not t1w_found:
-                LOGGER.warning('No T2w-space T1w found. Attempting T2w-only processing.')
+            if not t2wspace_t1w_found:
```
mattcieslak, line 289: I get confused because there is a T2-SPACE acquisition. This is referring to the bids entity, right?

mattcieslak (approved):
LGTM. It also got me thinking that these skeleton yamls could get added to simbids too. It might be useful to have a big curated set of skeletons in that package that we can use to test other packages

xcp_d/utils/bids.py
tsalo: Oh, as opposed to the sequence? Yeah it's the BIDS entity- so we're looking for `space-T2w_T1w` files, which could be created if the pipeline coregistered the secondary anatomical contrast to the primary anatomical contrast before running Freesurfer and normalization, AFAICT. I don't think fMRIPrep or Nibabies actually does that yet though, so this is more future-proofing.

PR #1472, opened 2025-07-14
Drop dir and run entities from single-run tasks

tsalo:
This duplicates the outputs, but it will make it easier to query the datasets.

### DIFF
--- a/xcp_d/workflows/base.py
+++ b/xcp_d/workflows/base.py
@@ -473,8 +473,13 @@ It is released under the [CC0](https://creativecommons.org/publicdomain/zero/1.0
         # Assuming TR is constant across runs for a given combination of entities.
         TR = _get_tr(nb.load(task_files[0]))
 
+        # We only "concatenate" if scans are named with a run or direction entity.
+        multirun_entity = get_entity(task_files[0], 'run') is not None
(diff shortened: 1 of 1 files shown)


xcp_d/workflows/bold/concatenation.py
```diff
@@ -501,3 +507,107 @@ def _combine_name(in_files):
...
+    ... ])
+    ['/path/to/sub-01_task-rest_bold.nii.gz']
+
+    >>> _combine_concat_name([
+    ...     ['/path/to/sub-01_task-rest_dir-AP_run-1_bold.nii.gz',
+    ...      '/path/to/sub-01_task-rest_dir-PA_run-2_bold.nii.gz'],
+    ... ])
+    ['/path/to/sub-01_task-rest_bold.nii.gz']
+
+    >>> _combine_concat_name([
```
mattcieslak, line 540: Are you sure you don't want dir-AP here

tsalo: We concatenate across the run and dir entities because of HCP. It may not always be the behavior people want, but I think folks will want to treat data acquired across multiple directions as the same more often than not.

PR #1473, opened 2025-07-14
Ingress BOLD brain mask in ABCD-BIDS and HCP processing

tsalo:
Closes none, but addresses a bug identified by @rosemccollum. Basically, in 0.10.7 we started using a brain mask NIfTI in the CIFTI workflow in order to mask out non-brain voxels in the executive summary (see #1425), but I didn't add a brain mask to the converted ABCD/HCP data.


tsalo:
I'm going to merge and check with rae before making a patch release.

PR #1487, opened 2025-08-12
Use MNIInfant transforms from templateflow

### LINKED ISSUE #1317: Raise an error when using nibabies inputs + nifti processing + atlases
We currently rely on a simple affine transform to get NIfTI atlases from MNI152NLin6Asym to MNIInfant, which is pretty terrible. That shouldn't affect HBCD, since HBCD uses CIFTI processing.

tsalo:
Closes #1317.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,7 @@ orbs:
 .dockersetup:
   &dockersetup
   docker:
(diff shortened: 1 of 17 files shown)


tsalo:
The problem is an outdated templateflow cache. Unfortunately, trying to clear it within the test isn't working, so I think I need to update xcpd_build.

PR #1496, opened 2025-08-20
Add hash entity to outputs

### LINKED ISSUE #1495: Incorporate hash entity into filenames
This is tied to https://github.com/nipreps/nibabies/issues/476.

1. Add parameter to CLI (possibly `--output-layout <bids|multiverse>`, based on fMRIPrep). If `output-layout` is `multiverse`, do the following.
2. Calculate a hash from a subset of parameters in the config object. If the input dataset has a hash value in its `dataset_description.json`, incorporate that into XCP-D's hash (e.g., `nibabieshash+xcpdhash`).
3. Add hash to the `dataset_description.json`.
4. Add hash as entity to output filenames.
5. I'm not sure if we need to include the hash in the column names of tabular files or not. Waiting on guidance from the DCAN team on this.

tsalo:
Closes #1495.

- Add `--output-layout` parameter to CLI, with default of `bids` (current behavior) and new option (`multiverse`). When multiverse layout is selected, the following happens:
    - A new folder within `output_dir` named `xcp_d-<hash>` (where `<hash>` is a hash based on a subset of the parameters selected by the user) is used instead of `output_dir`.
    - A new `hash-<label>` entity is added to all output filenames.
    - The `hash-<label>` entity is added to all column names in tabular data (to be implemented).
    - The hash is checked against an existing hash if a dataset exists in the output directory. If the hash doesn't match, then an error is raised.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -283,7 +283,7 @@ jobs:
             mkdir /src/coverage
             mv /src/xcp_d/.coverage /src/coverage/.coverage.ds001419_nifti
             # remove nifti files before uploading artifacts
-            # find /src/xcp_d/.circleci/out/ -name "*.nii.gz" -type f -delete
+            find /src/xcp_d/.circleci/out/ -name "*.nii.gz" -type f -delete
       - persist_to_workspace:
           root: /src/coverage/
           paths:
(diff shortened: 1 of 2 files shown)


xcp_d/config.py
```diff
@@ -793,3 +798,88 @@ def to_filename(filename):
...
+    import json
+    from hashlib import blake2b
+    from pprint import pprint
+
+    if version is None:
+        from xcp_d import __version__ as version
+
+    # Load the preprocessing derivatives dataset description
+    pprint(conf)
+    dset_desc_path = Path(conf['execution']['fmri_dir']) / 'dataset_description.json'
```
tsalo, line 870: I suppose technically we should use the hashes from all of the linked datasets, but I don't think the HBCD/ABCD folks care about hashes outside of the preprocessing pipeline.

PR #1565, opened 2026-01-30
Change smoothing surf from 'sphere' to 'midthickness'

### LINKED ISSUE #1564: Smoothing with fsLR spheres
Hello,

Looking at `init_resd_smoothing_wf`, I wanted to ask, what's the rationale behind running CiftiSmooth with the fsLR template spheres, rather than a template midthickness or session-specific midthickness (or sphere)?

I've seen this command used outside of XCP-D with the subject-specific fsLR-resampled midthicknesses, which is also what seems to be used in the docstring example inside of the CiftiSmooth interface.

Thanks!

joey-scanga:
Fix for #1564

Fix for #1564 which replaces spheres with midthicknesses when smoothing fsLR-resampled CIFTIs

### DIFF
--- a/xcp_d/workflows/bold/postprocessing.py
+++ b/xcp_d/workflows/bold/postprocessing.py
@@ -830,7 +830,7 @@ The denoised BOLD was then smoothed using *Connectome Workbench* with a Gaussian
                         hemi='R',
                         density='32k',
                         desc=None,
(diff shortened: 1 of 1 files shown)


tsalo (approved):
LGTM. Thanks for this!

PR #1159, opened 2024-05-01
Add custom censoring options

tsalo:
Closes #724 and closes #723.

To do:

- [ ] Expand testing of both argument parsing and actual censoring
- [ ] Plot DVARS in CensorPlot
- [ ] Make sure the new method is documented thoroughly
- [ ] Cover the new arguments in integration tests

## Changes proposed in this pull request

- Add new parameters to support more customized censoring:
    - `--fd-thresh` can accept a second value, which applies additional FD-based censoring after interpolation+filtering+denoising+recensoring.
    - `--dvars-thresh`, which accepts one or two values, to apply a DVARS-based threshold when identifying outliers.
    - `--censor-before` and `--censor-after`, which both accept one or two values to censor volumes before/after outlier volumes identified by `--fd-thresh` and `--dvars-thresh`.
    - `--censor-between`, which accepts one or two values to censor contiguous sequences of non-loutlier volumes between flagged outlier volumes, as long as the sequence is shorter than or equal to the `censor-between` value.
- Expand the outliers TSV file to have 6 columns:
    - `framewise_displacement`: Volumes with FD greater than the first value of `--fd-thresh`
    - `dvars`: Volumes with DVARS greater than the first value of `--dvars-thresh`
    - `denoising`: Volumes flagged by the `framewise_displacement` or `dvars` columns, expanded using the first values of `--censor-before`, `--censor-after`, and `--censor-between`
    - `framewise_displacement_interpolation`: Volumes with FD greater than the second value of `--fd-thresh`
    - `dvars_interpolation`: Volumes with DVARS greater than the second value of `--dvars-thresh`
    - `interpolation`: Volumes flagged by any of the previous columns, expanded using the second values of `--censor-before`, `--censor-after`, and `--censor-between`.
- Split up the `GenerateConfounds` interface into `ModifyConfounds` (motion filtering and FD recalculation), `GenerateTemporalMask` (outlier detection using FD and/or DVARS confounds), and `GenerateDesignMatrix` (select confounds, combine with custom confounds, and orthogonalize noise regressors w.r.t. signal regressors).

Issue #724, opened 2022-12-15
Customizing motion scrubbing logic


Bramdiamond:
## Summary
Target volumes that exceed FD **or** DVARS thresholds. It would be nice to have the option to customize this boolean logic.

## Next steps
For modified_data.py, add compute_dvars() and generate_mask() to include dvars. xcp_d options might include --dvars-thresh, --despike-logic <and, or>.

Additionally, xcp_p should be able to read FD/DVARS from fMRIPrep output.

tsalo:
We discussed this in a recent meeting and agreed that it's a good idea. I like `--dvars-thresh`. Would you be willing to open a PR? If not, it might be a little bit before I'm able to implement this.

Bramdiamond:
I've never submitted a PR, but I can take a look!

tsalo:
We'll want to avoid extreme flexibility, so it would be great to limit this option to specific, established strategies. @Bramdiamond are you familiar with any citable approaches?

tsalo:
I'm thinking of something like this:

```
--fd-thresh <FLOAT> [<FLOAT>]
--dvars-thresh <FLOAT> [<FLOAT>]
--censor-config [before=<INT>] [after=<INT>] [contiguous=<INT>]
```

I don't love how complicated it ends up being, but here's my thinking.

1. The DCAN team prefers to fit the nuisance regression and perform the interpolation using one FD threshold, then censor the denoised, interpolated time series using a second, stricter threshold. Allowing a second, optional value for `--fd-thresh` and the new `--dvars-thresh` would support this behavior.
    - I believe the primary goal of this approach is to address the spreading of interpolated data to neighboring volumes by the temporal filter, but I will follow up with the DCAN team to make sure I'm understanding it correctly.
3. Users would probably want to censor some number of volumes before and after the high-motion volumes, hence `before` and `after`.
4. Users may want to censor short sets of low-motion volumes between censored volumes, hence `contiguous`. In this case, any runs of contiguous low-motion volumes shorter than the value would be censored.

One thing I'm not sure about is where the censoring should be done. On the one hand, we want to ensure that the nuisance regression is not adversely impacted by noisy volumes, which is what `--fd-thresh` is meant to address. On the other hand, interpolation + bandpass filtering + censoring with the original threshold will result in interpolated signal spread to neighboring volumes, so we need to apply some additional censoring unrelated to the nuisance regression.

Bramdiamond:
Hi Taylor,

[...]

PR #1628, opened 2026-07-30
Add `--censor-between` parameter to censor short sets of nonoutliers between outliers

tsalo:
Closes none, but relates to #724 and #1159. `--censor-between N` censors any contiguous segment of non-outlier volumes of length N or fewer. Default is 0 in every mode (for now).

- New function `censor_between_outliers`.
- New interface `ExpandTemporalMask`.
- Temporal mask will now have two new columns: `censor_between` and `denoising` (the union of `framewise_displacement` and `censor_between`.


tsalo:
@lucimoore @madisoth does this do what you need?

madisoth:
Just got around to testing this and it just needs to also have the censoring applied to the masks in the ABCC QC hdf5 file (and the remaining frames, mean FD, etc. to reflect the extra censoring) -- the rest looks all good to me.

tsalo:
@madisoth I made the change you requested and have released 26.2.0rc1. Can you test it out to make sure it does what you want before I release 26.2.0?

madisoth:
looks good, thanks!

